PR #4, opened 2018-04-23
framework update

andrewxhill:


asutula:
Cool dude, but I think we should leave `YellowBox` warnings enabled, but just disable specific ones using the way here https://facebook.github.io/react-native/docs/debugging.html#warnings

asutula:
Something like `YellowBox.ignoreWarnings(['Warning: isMounted(...) is deprecated'])`

asutula:
And I guess that would go in `index.js`

PR #6, opened 2018-04-23
Image upload error redux

asutula:
This should make the errors appear in our redux logger. Also cleaned up redux to get rid of unused code.


andrewxhill:
ah, nice cleanup. last time i went into that list of methods i cringed a bit.

PR #10, opened 2018-04-24
Photos task saga + Redux & logging improvements

asutula:
This makes less of the image processing bypass Redux and our logging. Location updates, foregrounding events, and background tasks now trigger redux actions (and are logged to App Center), and the photo querying and image add now happen in a Saga. All around this leads to better logging and error handling. 

The App Center payload is now standardized to look like `{ description: string, error: string }`, and there is a mapping function that takes any redux action and converts it to that standard payload.


andrewxhill:
very nice, getting pretty tight again

PR #12, opened 2018-04-25
forces update at end of onboarding success... may dup with location

andrewxhill:


andrewxhill:
location update triggers after onboarding, which then triggers the photos. i don't think in the end that this is needed, it just duplicates the photos saga

asutula:
yea i agree, was wanting to investigate before merging this, so cool

asutula:
so is the potential issue just the node taking a long time to start up?

PR #17, opened 2018-04-26
Aaron/login screen

asutula:
This is a solid login screen with the most simple possible integration so far. Just want to get it merged and knock out additional features as new PRs. 

This uses some stubbed login and sign up functions, and it's only displayed at the beginning of the onboard sequence (no handling of subsequently invalidated sessions yet).


sanderpick:
don't have my head around this RN setup, but looks good to me. i don't think @andrewxhill will even open the diff, so i'll merge it

PR #20, opened 2018-04-27
Migration to RN upload library

andrewxhill:


andrewxhill:
okay, this is close. i just don't like the way i'm using an outer var in a listener. see comment. maybe okay

UploadTaskNativeModule.js
```diff
@@ -1,18 +1,35 @@
 //  Created by react-native-create-bridge
```
asutula: Ok, so this file is (was) our native module JS file. We should be able to remove this file and use `react-native-background-upload`'s `Upload` directly. You're basically just wrapping in here.

asutula:
I don't see a comment?

UploadTaskNativeModule.js
asutula: I see you imported it in `TextileSagas.js` like you were going to use it, but I don't see any other change in that file.

App/Redux/TextileRedux.js
```diff
@@ -54,11 +59,22 @@ export const handleImageAdded = (state, {image, remotePayloadPath}) => {
   return state.merge({ images: { items } })
 }
 
+export const handleImageSubmittedForUpload = (state, {uploadId}) => {
+  const existingItems = state.images.items ? state.images.items : []
+  const items = existingItems.map(item => {
+    if (item.uploadId === uploadId) {
```
asutula: If I understand this, this reducer is meant to update the image data for the image that has just been submitted for upload, updating it's state with the `uploadId`. So a couple issues... 

The image should be one of the action arguments (you're only extracting `uploadId` above. You'll have to find the image to update using `remotePayloadPath`, the same way we do below in `handleImageProgress`. Once you find it, you want to add a new property to it, the `uploadId`. I don't see that happening anywhere.

App/Redux/TextileRedux.js
```diff
@@ -67,18 +83,28 @@ export const handleImageProgress = (state, {data}) => {
...
 
 export const handleImageUploadComplete = (state, {data}) => {
-  const {file} = data
-  // Remove the item from Redux storage
-  const items = state.images.items.filter(item => item.remotePayloadPath !== file)
+  const { uploadId } = data
+  const existingItems = state.images.items ? state.images.items : []
+  const items = existingItems.map(item => {
+    if (item.uploadId === uploadId) {
+      return {...item, state: 'cleanup'}
```
asutula: Why is this additional `cleanup` state necessary?

App/Redux/TextileRedux.js
```diff
@@ -67,18 +83,28 @@ export const handleImageProgress = (state, {data}) => {
...
 
 export const handleImageUploadError = (state, {data}) => {
-  const {file, error} = data
+  const { error, uploadId } = data
   const existingItems = state.images.items ? state.images.items : []
   const items = existingItems.map(item => {
-    if (item.remotePayloadPath === file) {
-      return {...item, state: 'error', error: error.message}
+    if (item.uploadId === uploadId) {
+      return {...item, state: 'error', error: error}
```
asutula: I've seen that actual `Error` objects don't persist properly in Redux Persist... Better to just store `error.message` like was there before.

App/Sagas/TextileSagas.js
```diff
@@ -130,13 +130,18 @@ export function * photosTask (action) {
...
+        Upload.startUpload,
+        {
+          path: multipartData.payloadPath,
+          url: 'https://ipfs.textile.io/api/v0/add?wrap-with-directory=true',
+          method: 'POST',
+          type: 'multipart',
+          field: multipartData.boundary
+        }
       )
+      photo['uploadId'] = uploadId
```
asutula: This doesn't do anything, and you can't/shouldn't update any state in a saga anyways.

App/Sagas/TextileSagas.js
```diff
@@ -130,13 +130,18 @@ export function * photosTask (action) {
...
+      yield put(TextileActions.imageSubmittedForUpload(uploadId))
```
asutula: I think this action should take two args, the photo and the upload id. The reducer that handles it will be responsible for updating the image state to include the id. See my comments in the reducer code.

App/Sagas/TextileSagas.js
andrewxhill: ah, was replicating the photo['hash'] = something a few above. will drop that too

App/Redux/TextileRedux.js
andrewxhill: not necessary, i just figured if we get around to completing all the error handling, there is the potential error of an image uploading correctly but failing to cleanup it's multipart data

App/Sagas/TextileSagas.js
```diff
@@ -146,6 +151,13 @@ export function * photosTask (action) {
 }
 
 export function * removePayloadFile ({data}) {
-  const {file} = data
-  yield call(RNFS.unlink, file)
+  const { uploadId } = data
+  const items = yield select(TextileSelectors.items)
```
asutula: Yea seems like this would be cleaner if there were no `cleanup` state, and this saga was fired after an image was successfully uploaded (like it was configured before). The Selector in here could just return the file path for the given upload id. Then the complexity of filtering state looking for the image would be in the selector and this could remain simple like: 
```
const { uploadId } = data
const file = yield select(TextileSelectors.fileForUploadId, uploadId)
yield call(RNFS.unlink, file)
```

App/Redux/TextileRedux.js
andrewxhill: oop, good catch meant to revert this

App/Sagas/TextileSagas.js
andrewxhill: it will fire before the original handleImageUploadComplete, so then there wasn't a record of the path in the items that it could get

App/Redux/TextileRedux.js
```diff
@@ -36,10 +38,13 @@ export const INITIAL_STATE = Immutable({
   devices: []
 })
 
-/* ------------- Selectors ------------- */
 
+/* ------------- Selectors ------------- */
 export const TextileSelectors = {
   // TODO: Add more selectors here as we learn how they are used
+  items: state => {
```
asutula: Was thinking this could be a function like: ```
getPathForUploadId: (state, uploadId) => {
  // Do all the filtering of the state in here to find the image based on the uploadId and return the file path
}
```

andrewxhill:
okay, this gets all of it except that 'cleanup' step. which if i remove, the record of the upload is dropped before i have a chance to remove the upload file from disk

PR #23, opened 2018-04-28
Android back to working state

andrewxhill:


andrewxhill:
okay, i'd like to go ahead and get this merged. i can then work on those styling things in a follow-up

TextileIPFSNativeModule.js
```diff
@@ -10,28 +10,48 @@ type MultipartData = {
 
 export default {
   createNodeWithDataDir: async function (dataDir: string): boolean {
-    const success = await TextileIPFS.createNodeWithDataDir(dataDir)
-    return success
+    try {
+      const success = await TextileIPFS.createNodeWithDataDir(dataDir)
+      return success
+    } catch (e) {
```
asutula: I don't think we want to catch any of these errors... they need to bubble up so the rest of the app works correctly

andrewxhill: only thing is that i was getting some error at one point that wasn't giving me any meaningful information in the old setup. at least this gave me the log. maybe there is a proper crash handler i could put in instead of just .log?

asutula: I think the errors should be handled and dispatched as Redux actions, in which case you could see them in Reactron. This all happens in our sagas. Check out how all the functions in there that interact with the IPFS native module are using `try/catch` and dispatching an action with the error in the case of any error, for example:
```
try {
    const success = yield call(IPFS.createNodeWithDataDir, path)
    if (success) {
      yield put(IpfsNodeActions.createNodeSuccess())
    } else {
      yield put(IpfsNodeActions.createNodeFailure(new Error('Failed creating node, but no error was thrown - Should not happen')))
    }
  } catch (error) {
    yield put(IpfsNodeActions.createNodeFailure(error))
  }
```

asutula: I think the only other place we interact with the IPFS native module is in the photo grid and photo viewer, and that's just a temporary hack until we can load images via URI. If those are the places causing problems you need to log, I think you could put `try/catch/log` code directly into those files.

carsonfarmer: Those will be gone in short order anyway... Take a look here:
https://github.com/textileio/textile-mobile/pull/28

andrewxhill: okay, fixec

PR #27, opened 2018-04-30
Refactor out non-UI event listeners, tabbed UI

asutula:
We have a bunch of event listeners that are used to trigger photo queries and update state of photo data. They were all created as part of `TextilePhotos.js`, which is really just the UI for viewing the photo grid. This isn't good design, and it makes the photo grid impossible to re-use. 

This PR does two things:

1. Move all event listeners and redux triggers to a wrapper component, keeping `TextilePhotos.js` more general use and more reusable.
2. Create a tabbed interface for now that we can plug the first cut of shared photos into. It will look something like:
screen shot 2018-04-30 at 15 24 32
That isn't the end design or anything, just a place to start for early testers.

### DIFF
--- a/App/Containers/OnboardingCheck.js
+++ b/App/Containers/OnboardingCheck.js
@@ -7,7 +7,7 @@ class OnboardingCheck extends Component {
     super(props)
     // This will switch to the App screen or Auth screen and this loading
     // screen will be unmounted and thrown away.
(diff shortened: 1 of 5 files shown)


asutula:
The idea behind making the photo grid reusable, is to quickly implement the shared photo UI by using it.

PR #28, opened 2018-05-01
adds https support to mobile ios

carsonfarmer:
adds support for a mobile https server for pinned assets. still a work in progress in that cookies need to be implemented, and we still need to decide what to include in terms of metadata on the photo 'zoom' view (which was also added to this pr).


ios/RCTHTTPRequestHandler+SelfSigned.m
```diff
@@ -0,0 +1,15 @@
+//
```
carsonfarmer: This avoids us having to modify any react-native libs directly ourselves

ios/TextilePhotos.xcodeproj/project.pbxproj
```diff
@@ -2252,6 +2122,8 @@
 					"$(SRCROOT)/../node_modules/appcenter-analytics/ios/AppCenterReactNativeAnalytics",
 					"$(SRCROOT)/../node_modules/appcenter-crashes/ios/AppCenterReactNativeCrashes",
 					"$(SRCROOT)/../node_modules/react-native-config/ios/ReactNativeConfig",
+					"$(SRCROOT)/../node_modules/react-native/Libraries/Network/**",
```
carsonfarmer: These two items are needed to access react-native headers in our code above

carsonfarmer:
Should be ready for review @asutula. Aside from the cookie stuff (though I have included the cookie lib here).

App/Containers/PhotoViewerScreen.js
```diff
@@ -32,55 +50,41 @@ class PhotoViewerScreen extends React.PureComponent {
...
+      </View>
     )
   }
 
   render () {
+    const item = this.props.navigation.state.params.item
     return (
       <View style={{flex: 1}}>
         <StatusBar hidden />
-        <Gallery
```
asutula: Guess I'm a little confused why you removed the `Gallery` component. So now you just just view a single image, but you can't swipe to the next/previous images? And there's probably no zooming either?

App/Containers/TextilePhotos.js
```diff
@@ -50,13 +50,13 @@ class TextilePhotos extends React.PureComponent {
...
     }
-    const imageData = IPFS.syncGetPhotoData(item.image.hash + '/thumb')
+    // TODO: Programatically determine gateway port rather than hard-coding
     return (
       <TouchableOpacity onPress={onPress} >
         <View style={styles.item}>
           <View style={styles.itemBackgroundContainer}>
             <Image
-              source={{uri: 'data:image/jpeg;base64,' + imageData}}
+              source={{uri: 'https://localhost:9080/ipfs/' + item.image.hash + '/thumb'}}
```
asutula: Nice!

App/Containers/PhotoViewerScreen.js
```diff
@@ -1,19 +1,38 @@
...
-// Add Actions - replace 'Your' with whatever your reducer is called :)
-// import YourActions from '../Redux/YourRedux'
 
 // Styles
 import styles from './Styles/PhotoViewerScreenStyle'
 import {buttonColor1} from "./Styles/InfoViewStyle";
 
 class PhotoViewerScreen extends React.PureComponent {
 
+  // componentWillMount() {
```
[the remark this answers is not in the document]
carsonfarmer: yeh i'm not actually doing anything with this... its just in case we wanted to query the metadata via the hash, but its commented out for now

App/Containers/PhotoViewerScreen.js
carsonfarmer: i didn't think of those other use cases actually :P as per our dicsussion on slack, i'll revert this

App/Containers/PhotoViewerScreen.js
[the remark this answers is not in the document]
asutula: Cool. Yea it's probably better to use the native module JS api anyways. I really think of the gateway as just an efficient way to get images.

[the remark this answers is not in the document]
carsonfarmer: File has been reverted to previous state

App/Containers/PhotoViewerScreen.js
carsonfarmer: file has been reverted to a previous state

App/Containers/TextilePhotos.js
carsonfarmer: :)

asutula:
You want to deal with these conflicts? For the pbxproj conflict, please take the version from master, and then you can re-add RCTHTTPRequestHandler+SelfSigned.m to the project. That will be way more simple than trying anything else.

PR #33, opened 2018-05-02
Just connects signin methods

andrewxhill:


andrewxhill:
Okay, just tested these with real hashes... all good. Really simple. Just fails, didn't hack any error display into the form yet, wanted to leave that to you aaron.

PR #36, opened 2018-05-03
Removes manual gateway start starts gateway in node

carsonfarmer:
Should be GTG @asutula


carsonfarmer:
actually, hold off one sec, i'm testing on device to be 100% sure

PR #38, opened 2018-05-03
carson/mobile cookies for ios

carsonfarmer:


carsonfarmer:
Huh, well this is an issue:
https://github.com/joeferraro/react-native-cookies/issues/84
I might have to update the lib I'm using for writing cookies. Should be easy enough...

andrewxhill:
can we launch without this or is the a requirement for the secure calls?

carsonfarmer:
Right now, we have https working fully. But without the cookie stuff, someone with access to localhost could have requests and get photos. 
I think if I can do a platform check from RN, we can use this lib for now. Then it’s just a matter of exposing the gateway password to RN, which I think is easy enough.

andrewxhill:
yeah, `if (Platform.OS === 'android')` is all you want.

carsonfarmer:
ok, this is finally good to go... works on ios. probably won't work on android given a) the https issues and b) the client-side cookie adding has not been tested properly

PR #42, opened 2018-05-04
adds framework from latest textile-go master with cookies off

carsonfarmer:
just in case you wanna test without cookies (in case that's also an issue on android)


carsonfarmer:
ready for you to merge into your branch if you want it andrew

PR #46, opened 2018-05-05
Stop node on inactive state, clean restart, update frameworks

sanderpick:
The issue here was that after a backgrounding, the node would go offline for some reason. This is only seen in a release build... everything is fine when debugging.

Took me awhile to get this right... back and forth between here and go layer. Basically, we just need to stop the node when it goes to `inactive` state. https://github.com/textileio/textile-go/pull/87 lets the node cleanly restart.

### DIFF
--- a/App/Sagas/TextileSagas.js
+++ b/App/Sagas/TextileSagas.js
@@ -115,9 +115,20 @@ export function * startNode () {
   }
 }
 
+export function * stopNode ({newState}) {
+  if (newState !== 'inactive') {
+    return
+  }
+  try {
+    yield call(IPFS.stopNode)
+  } catch (error) {
+    yield put(IpfsNodeActions.stopNodeFailure(error))
+  }
+}
+
 export function * getPhotoHashes ({thread}) {
   try {
-    const photoData = yield call(IPFS.getPhotos, null, 100000, thread)
+    const photoData = yield call(IPFS.getPhotos, null, -1, thread)
     yield put(IpfsNodeActions.getPhotoHashesSuccess(thread, photoData))
   } catch (error) {
     yield put(IpfsNodeActions.getPhotoHashesFailure(thread, error))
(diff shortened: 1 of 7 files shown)


sanderpick:
Also updates ios and android frameworks to what's in https://github.com/textileio/textile-go/pull/87.

App/Sagas/TextileSagas.js
```diff
@@ -115,9 +115,20 @@ export function * startNode () {
...
+    yield call(IPFS.stopNode)
+  } catch (error) {
+    yield put(IpfsNodeActions.stopNodeFailure(error))
+  }
+}
+
 export function * getPhotoHashes ({thread}) {
   try {
-    const photoData = yield call(IPFS.getPhotos, null, 100000, thread)
+    const photoData = yield call(IPFS.getPhotos, null, -1, thread)
```
sanderpick: "unlimited"

App/Sagas/TextileSagas.js
```diff
@@ -155,14 +166,6 @@ export function * shareImage ({thread, hash}) {
   }
 }
 
-export function * refreshThreads ({newState}) {
```
sanderpick: removed this and included in `photosTask` since it was causing a timing issue, even if the hookup in `index.js` was places after the one for `photosTask`, I think because the latter happens in a background task? Anyway, this just needs to happen after the node starts. Previously, this wasn't an issue cause the active state doesn't seem to be triggered on the first load... and for subsequent occurrences, although it was not working, the node was already "started".

TextileIPFSNativeModule.js
```diff
@@ -60,7 +60,7 @@ export default {
   },
 
   getPhotos: async function (offset: ?string, limit: number, thread: string): string {
-    const result = await TextileIPFS.getPhotos(offset, limit, thread)
+    const result = await TextileIPFS.getPhotos(offset || "", limit, thread)
```
sanderpick: saw a couple error related to receiving `null`s in go layer here, but maybe misleading - anyway, the empty string is expected in the go lib.

sanderpick:
@asutula the app setup rules! a pleasure to work on.

App/Sagas/TextileSagas.js
asutula: Nice

asutula:
Cool. So I guess the foreground event causes the node to start back up because `photosTask()` saga calls start up? That's fine for now, might be good to make that a little more clear in the future.

PR #49, opened 2018-05-07
Simpler HTTP gateway with rotating random passwords for all resources

carsonfarmer:
Passwords are invalidated and randomized after each time they are requested. RN does some useful caching by default, to avoid getting stale image urls I think (which is good).

### DIFF
--- a/App/Containers/PhotoViewerScreen.js
+++ b/App/Containers/PhotoViewerScreen.js
@@ -62,8 +62,8 @@ class PhotoViewerScreen extends React.PureComponent {
 
 const mapStateToProps = (state, ownProps) => {
   const hashes = state.ipfs.threads[ownProps.navigation.state.params.thread].hashes
-  const imageData = hashes.map(hash => {
-    return { hash, source: { uri: 'https://localhost:9080/ipfs/' + hash + '/photo' } }
+  const imageData = hashes.paths.map((path, idx) => {
+    return { hash: hashes.hashes[idx], source: { uri: 'http://' + path + '/photo' } }
   })
   return {
     imageData,
--- a/App/Containers/PhotoViewerScreen.js
+++ b/App/Containers/PhotoViewerScreen.js
@@ -17,20 +17,20 @@ class PhotoViewerScreen extends React.PureComponent {
     this.props.screenProps.dismiss()
   }
 
-  // renderImage(row) {
-  //   var item = row.image.data
-  //   console.warn(item)
-  //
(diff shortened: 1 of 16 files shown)


App/Containers/PhotoViewerScreen.js
```diff
@@ -62,8 +62,8 @@ class PhotoViewerScreen extends React.PureComponent {
 
 const mapStateToProps = (state, ownProps) => {
   const hashes = state.ipfs.threads[ownProps.navigation.state.params.thread].hashes
-  const imageData = hashes.map(hash => {
-    return { hash, source: { uri: 'https://localhost:9080/ipfs/' + hash + '/photo' } }
+  const imageData = hashes.paths.map((path, idx) => {
+    return { hash: hashes.hashes[idx], source: { uri: 'http://' + path + '/photo' } }
```
carsonfarmer: this could just be parsed from the uri, rather than storing a separate list

App/Redux/IpfsNodeRedux.js
```diff
@@ -31,12 +31,18 @@ export const INITIAL_STATE = Immutable({
   threads: {
     default: {
       querying: false,
-      hashes: [],
+      hashes: {
```
asutula: I guess this would be cleaner if `hashes` was an array of objects, those objects looking like:
```
{ 
  hash: string,
  uri: string
```
Also seems like the 'path' should just be the uri to the ipfs directory, including the protocol (http, etc)

ios/NativeModules/TextileIPFS.m
```diff
@@ -96,8 +96,14 @@ - (dispatch_queue_t)methodQueue {
...
   id json = [NSJSONSerialization JSONObjectWithData:data options:0 error:nil];
   NSArray *hashes = [json objectForKey:@"hashes"];
+  NSArray *paths = [json objectForKey:@"paths"];
+  // TODO: Could we just return the json directly?
   if (hashes) {
-    resolve(hashes);
+    if (paths) {
+      resolve(@{ @"hashes": hashes, @"paths": paths });
+    } else {
+      resolve(@{ @"hashes": hashes });
```
asutula: This should never happen and wouldn't make any sense if it did. Related to my comment about the data structure being returned having separate hashes and paths properties, when those things are actually related and should be grouped into an object.

carsonfarmer: I’ll fix, don’t merge.

PR #50, opened 2018-05-08
Sharing Dialog

asutula:


App/Containers/Comment.js
```diff
@@ -0,0 +1,49 @@
+import React, { Component } from 'react'
```
asutula: This UI to enter a photo comment isn't being used yet, just leaving it in place for when we're ready.

App/Redux/UIRedux.js
```diff
@@ -0,0 +1,40 @@
+import { createReducer, createActions } from 'reduxsauce'
```
asutula: Created this redux config for tracking user interaction flows, like sharing an image. It isn't persistent.

PR #51, opened 2018-05-08
rough pass

andrewxhill:


andrewxhill:
ended up reverting a lot of your RN changes carson, so pointing this at master now to just skip your PR

andrewxhill:
alright guys, ios is in a good working state here. worth reviewing and testing. 

android works... but other issues with rendering and now sharing

TextileIPFSNativeModule.js
```diff
@@ -69,6 +79,26 @@ export default {
...
+    await TextileIPFS.setHashToken(hash, token)
+  },
+
+  getHashRequest: function (hash: string, path: string): HashRequest {
+    let token = ''
+    const chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
+    for (let i = 0; i < 48; i++) {
+      token += chars.charAt(Math.floor(Math.random() * chars.length))
+    }
+    this.registerToken(hash, token)
```
asutula: Calling a async function here, no guarantee it will complete before the request returned from this function is made. I think as this is currently set up, we're basically hoping the async call will finish in time.

andrewxhill: yep, 100% the intention :)

andrewxhill: haha

App/Containers/PhotoViewerScreen.js
```diff
@@ -23,6 +25,15 @@ class PhotoViewerScreen extends React.PureComponent {
     this.props.screenProps.dismiss()
   }
 
+  renderImage(props, dims) {
```
asutula: There was a version of this from back when we were rendering base64 data that passed more of the props/dims into the Image being returned... maybe a good idea to resurrect some of that code?

App/Containers/PhotoViewerScreen.js
```diff
@@ -85,8 +97,10 @@ class PhotoViewerScreen extends React.PureComponent {
 const mapStateToProps = (state, ownProps) => {
   const hashes = state.ipfs.threads[ownProps.navigation.state.params.thread].hashes
   const imageData = hashes.map(hash => {
-    return { hash, source: { uri: 'https://localhost:9080/ipfs/' + hash + '/photo' } }
+    // todo, try source here again
+    return { hash, source: require('../Images/TextileHeader.png')}
```
asutula: You can just put in a fake url for source so it doesn't actually try to load this image resource when we're not even using it.

asutula: Just put `file://foo.jpg` or something.

TextileIPFSNativeModule.js
andrewxhill: actually, just realized. in ios i can just make it sync, keep it sync in android for now

App/Containers/PhotoViewerScreen.js
andrewxhill: taking a look now

App/Containers/PhotoViewerScreen.js
andrewxhill: done

andrewxhill: any invalid placeholder messes it up. this was simply to avoid this warning,

screen shot 2018-05-08 at 5 55 38 pm

maybe if we don't care about that warn i can just ignore it

asutula: hmm, i think as long as it's a valid source object (`{uri: string}`) it should be fine. That's also something you can see in the base64 commits

andrewxhill: so it does... fixed

TextileIPFSNativeModule.js
andrewxhill: done

asutula:
Ok so, is this the solution we want, or should we test my branch on Android, using potential fixes to the cookie based approach?

andrewxhill:
alright, this is pointed at your branc, aaron/nits because I had merged that in here. java sign-in is still broke from what i see, but i can test again later maybe it's my brain that is broke. Either way, I'd vote to get these two branches in asap and then fix from there.

it's using an Image Container to handle loading the images asyc, called AsyncImage.

App/Components/AsyncImage.js
```diff
@@ -0,0 +1,54 @@
+import React, { Component } from 'react'
+import { View, Image } from 'react-native'
+import IPFS from '../../TextileIPFSNativeModule'
+
+export default class AsyncImage extends React.PureComponent {
```
asutula: This is cool, really simplifies dealing with the async image source. We could make it more general use by having one of the props just be an async function that returns the image `source`.

andrewxhill:
okay, reverted that comment out for startup, so we can fix startup stuff in follow-up. i think this should be merged so we can all try out the performance of the asyncimage stuff

carsonfarmer:
Yeh nice @andrewxhill. Agreed, let's get this merge in. My tests show it working pretty consistently this morning on both OSes.

PR #52, opened 2018-05-09
Various improvements

asutula:
Update to Mobile.framework that starts up faster
Clean up sagas
Block UI until node is started

### DIFF
--- a/App/Redux/IpfsNodeRedux.js
+++ b/App/Redux/IpfsNodeRedux.js
@@ -4,10 +4,7 @@ import Immutable from 'seamless-immutable'
 /* ------------- Types and Action Creators ------------- */
 
 const { Types, Creators } = createActions({
-  createNodeRequest: ['path'],
-  createNodeSuccess: null,
-  createNodeFailure: ['error'],
-  startNodeRequest: null,
+  startNodeRequest: ['path'],
   startNodeSuccess: null,
   startNodeFailure: ['error'],
   stopNodeRequest: null,
@@ -25,7 +22,7 @@ export default Creators
 
 export const INITIAL_STATE = Immutable({
   nodeState: {
-    state: 'undefined', // | creating | stopped | starting | started | stopping
+    state: 'stopped', // | starting | started | stopping
     error: null
   },
   threads: {
@@ -48,12 +45,6 @@ export const IpfsNodeSelectors = {
 }
 
 /* ------------- Reducers ------------- */
-
-export const creatingNode = state =>
-  state.merge({...state, nodeState: {state: 'creating', error: null}})
-
-export const nodeCreated = state =>
-  state.merge({...state, nodeState: {state: 'stopped', error: null}})
 
 export const nodeStarting = state =>
   state.merge({...state, nodeState: {state: 'starting', error: null}})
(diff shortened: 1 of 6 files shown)


App/Redux/IpfsNodeRedux.js
```diff
@@ -4,10 +4,7 @@ import Immutable from 'seamless-immutable'
 /* ------------- Types and Action Creators ------------- */
 
 const { Types, Creators } = createActions({
-  createNodeRequest: ['path'],
```
asutula: Creating and starting the node are one in the same, so just combine them into `start`

carsonfarmer:
Awesome. Will defs test this later!

App/Sagas/TextileSagas.js
```diff
@@ -71,59 +71,79 @@ export function * recoverPassword ({data}) {
...
+    yield * triggerStopNode()
   }
+  yield call(BackgroundTimer.stop)
 }
 
-export function * startNode () {
+export function * startNode ({path}) {
+  const onboarded = yield select(TextileSelectors.onboarded)
+  if (!onboarded) {
+    return
```
sanderpick: nice

App/Sagas/TextileSagas.js
```diff
@@ -71,59 +71,79 @@ export function * recoverPassword ({data}) {
...
+    yield call(BackgroundTimer.stop)
   }
 }
 
-export function * stopNode ({newState}) {
-  if (newState !== 'inactive') {
-    return
-  }
+export function * stopNode () {
+  yield call(BackgroundTimer.start)
```
sanderpick: just checking, can this be scheduled by the OS to happen at some later date, or does it necessarily happen right away? if the former case is true, hopefully it's canceled before the node starts again

App/Sagas/index.js
```diff
@@ -39,19 +39,25 @@ export default function * root () {
...
-
     takeEvery(IpfsNodeTypes.GET_PHOTO_HASHES_REQUEST, getPhotoHashes),
 
     takeEvery(UITypes.SHARE_PHOTO_REQUEST, shareImage),
 
-    takeEvery(TextileTypes.APP_STATE_CHANGE, photosTask),
-    takeEvery(TextileTypes.LOCATION_UPDATE, photosTask),
-    takeEvery(TextileTypes.BACKGROUND_TASK, photosTask),
+    // Actions that trigger starting/stopping the node
+    takeEvery(action => action.type === TextileTypes.APP_STATE_CHANGE && action.newState !== 'background', handleStateChange),
```
sanderpick: sweet

App/Sagas/TextileSagas.js
asutula: man, good question. it's on my list to understand how to compose all these flows in a way where things are canceled etc as needed, but that's some complicated stuff. this should happen right away, but it seems possible that the app could be foregrounded again or receive a location update before the node is done stopping. It's a complicated dependency graph of things

andrewxhill:
you ever see this on this branch?

screen shot 2018-05-08 at 6 35 31 pm

this is with a valid name/pass

asutula:
No, but don't think i've tried to log in, only sign up. Trying now.

asutula:
Log in worked fine on iOS

App/Sagas/TextileSagas.js
sanderpick: yeah, tricky stuff. good thing the node shuts down almost immediately

andrewxhill:
okay, but login is broke in android in this branch. 

just merged it into my branch and will try to fix

andrewxhill:
appears it doesn't start in android any more. calling a test here and get `isSignedIn() on null object reference`

andrewxhill:
seems that without the Onboarded flag the node doesn't start. But login happens before onboarding.

andrewxhill:
see my coments, pretty sure signin is broke in android. going to merge and we can fix from master

PR #54, opened 2018-05-09
Create node immediately, start in possibly immediately, def after onboarding

asutula:
Split out node creation redux so it can be created immediately and started later, supports sign up / log in.

Current setup, created the node as quickly as possible and then starts it if onboarding is complete. If onboarding is incomplete, it starts the node after onboarding.


andrewxhill:
nice, lgtm

PR #55, opened 2018-05-09
Cleans up a few of the image display things

andrewxhill:


App/Components/AsyncImage.js
```diff
@@ -20,35 +20,25 @@ export default class AsyncImage extends React.PureComponent {
...
-          < Image
-            source={this.state.source}
-            resizeMode={resizeMode}
-            style={style}
-          />
-        }
-
-        {!this.state.loaded &&
+    if (this.state.loaded) {
+      return (
```
andrewxhill: primarily wanted to remove the nested View

App/Components/PhotoGrid.js
```diff
@@ -34,35 +34,33 @@ export default class PhotoGrid extends React.PureComponent {
     return <MyCustomCell title={item.title} description={item.description} />
   *************************************************************/
   renderRow (row) {
-    const {item} = row
```
andrewxhill: just getting rid of unneeded const

andrewxhill: ah, makes sense

PR #56, opened 2018-05-10
Captions on shares and logging updates

carsonfarmer:
So when creating a `newNode` from RN, this PR means no code changes on the RN/Javascript side... and we just get the debug state from the OS... is this a nice dev UX, or would it be preferable to have more fine-grained control from RN? (edited)
I guess if we leave it up to RN/Javascript, we could more easily play around with different `logLevels`?
opinions?


App/Containers/PhotoViewerScreen.js
```diff
@@ -21,12 +21,11 @@ import styles from './Styles/PhotoViewerScreenStyle'
 import {buttonColor1} from "./Styles/InfoViewStyle";
 
 class PhotoViewerScreen extends React.PureComponent {
```
andrewxhill: just a bunch of changes to have `items: {hash, comment}` instead of `hashes:[]`

App/Redux/UIRedux.js
```diff
@@ -5,7 +5,7 @@ import Immutable from 'seamless-immutable'
 
 const { Types, Creators } = createActions({
   authorPhotoShare: ['hash'],
-  sharePhotoRequest: ['thread', 'hash'],
+  sharePhotoRequest: ['thread', 'hash', 'caption'],
```
andrewxhill: just realizing we should go with 'comment' standard

App/Sagas/TextileSagas.js
```diff
@@ -141,8 +143,19 @@ export function * stopNode () {
 
 export function * getPhotoHashes ({thread}) {
   try {
-    const photoData = yield call(IPFS.getPhotos, null, -1, thread)
-    yield put(IpfsNodeActions.getPhotoHashesSuccess(thread, photoData))
+    const hashes = yield call(IPFS.getPhotos, null, -1, thread)
+    let data = []
+    for (const hash of hashes) {
+      try {
+        const meta = yield call(IPFS.getHashData, hash, '/comment')
```
andrewxhill: just shoves the comments in with hashes right on first get. after all the rest works we can combine this in go

carsonfarmer: yup nice

ios/NativeModules/TextileIPFS.m
```diff
@@ -259,7 +259,7 @@ - (NSString *)_getHashRequest:(NSString *)hash error:(NSError**)error {
   return token;
 }
 
-- (NSString *)_getPhoto:(NSString *)hashPath error:(NSError**)error {
+- (NSString *)_getHashData:(NSString *)hashPath error:(NSError**)error {
```
carsonfarmer: as with all the others, i think this is a good move: renaming to be more generic

App/Containers/PhotoViewerScreen.js
```diff
@@ -116,7 +119,7 @@ const mapStateToProps = (state, ownProps) => {
 
 const mapDispatchToProps = (dispatch) => {
   return {
-    share: (hash) => { dispatch(IpfsActions.shareImageRequest('beta', hash)) },
+    share: (item) => { dispatch(IpfsActions.shareImageRequest('beta', item.hash)) },
```
asutula: This this is broken because above in `sharePressed` you're passing in `item.hash` and here you're trying to use `item.hash`.

andrewxhill: doh. where do you see it passing in a item.hash?

App/Containers/PhotoViewerScreen.js
```diff
@@ -39,8 +38,8 @@ class PhotoViewerScreen extends React.PureComponent {
 
   sharePressed () {
     const page = this.refs.gallery.currentPage
-    const hash = this.props.imageData[page].hash
-    this.props.authorShare(hash)
+    const item = this.props.imageData[page]
+    this.props.authorShare(item.hash)
```
asutula: here

andrewxhill: `authorShare` is different than `share` though no?

andrewxhill: just made them both take item and dispatch item.hash for consistency

App/Sagas/TextileSagas.js
```diff
@@ -141,8 +143,19 @@ export function * stopNode () {
...
+    const hashes = yield call(IPFS.getPhotos, null, -1, thread)
+    let data = []
+    for (const hash of hashes) {
+      try {
+        const meta = yield call(IPFS.getHashData, hash, '/caption')
+        const caption = JSON.parse(Buffer.from(meta, 'base64').toString('ascii'))
+        data.push({hash, caption})
+      } catch (err) {
+        // gracefully return an empty caption for now
+        data.push({ hash, caption: '' })
```
asutula: Seems like setting `null` here would be better so we can do more simple checks to decide whether to render the comment component:
`if (caption)` instead of `if (caption && caption.length > )`

TextileIPFSNativeModule.js
```diff
@@ -74,11 +74,15 @@ export default {
...
   },
 
-  getPhotoData: async function (path: string): string {
-    const result = await TextileIPFS.getPhotoData(path)
+  getHashData: async function (hash: string, path: string): string {
+    const result = await TextileIPFS.getHashData(hash + path)
     return result
   },
 
+  syncGetHashData: function (hash: string, path: string): string {
```
asutula: We should just remove all sync functions since they don't work for Android. Can deal with this in a later PR.

andrewxhill: yeah, nobody uses this anyway so can just drop now

App/Containers/PhotoViewerScreen.js
asutula: ah right, didn't notice the difference

PR #59, opened 2018-05-10
blazing fast shared photos

andrewxhill:


andrewxhill:
not sure which warning, but i don't think so. i was just trying to extract any computation from the render function

PR #62, opened 2018-05-10
Carson/simpler onboarding

carsonfarmer:
Not all that happy with this, but should be GTG if someone wants to make some comments and maybe merge. I'll try to find a nicer way to do the terms of service link

### DIFF
--- a/App/Navigation/OnboardingNavigation.js
+++ b/App/Navigation/OnboardingNavigation.js
@@ -1,6 +1,8 @@
-import { PermissionsAndroid, Platform } from 'react-native'
+import { PermissionsAndroid, Platform, Linking, Text } from 'react-native'
+import React from "react"
 import { StackNavigator } from 'react-navigation'
 import OnboardingScreen from '../Containers/OnboardingScreen'
+import { Colors } from '../../Themes'
 import LoginScreen from '../Containers/LoginScreen'
 import {getPhoto} from '../Services/PhotoUtils'
 import Actions from '../Redux/TextileRedux'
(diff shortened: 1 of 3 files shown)


App/Navigation/PhotosNavigation.js
```diff
@@ -112,7 +112,7 @@ const TabNav = TabNavigator(
     tabBarComponent: TabBarBottom,
     tabBarPosition: 'bottom',
     animationEnabled: false,
-    swipeEnabled: false
```
asutula: I had specifically set this to false. Any reason it needs to be true?

App/Navigation/OnboardingNavigation.js
```diff
@@ -1,78 +1,51 @@
...
-      navigate('Onboarding4', params6)
-    }
-  }
-}
-
-const params6 = {
   header: 'You\'re helping us improve.',
-  message: 'As a valued Beta Tester, we want to know how, when, and why you are using the app. We anonymously collect data, including crashes, screen interactions, and feedback. If you want to provide direct feedback, please feel free to email us with you thoughts, comments, or ideas.',
+  message: (
+    <View>
```
asutula: `OnboardingScreen` currently is wrapping the value if this `message` prop in a pair of `<Text>` tags. So in this case, it's rendering `<Text><View>...</View></Text>`, which might work, but is not right. I'd just update `OnboardingScreen` to expect that the value of message is a component without assuming what it is exactly. Know what I mean?

App/Navigation/PhotosNavigation.js
carsonfarmer: oh sorry, that was a mistake... was playing around and didn't set it back. will revert

carsonfarmer:
yup i'll update those two things and push @asutula

App/Navigation/OnboardingNavigation.js
carsonfarmer: fixed

carsonfarmer:
GTG @asutula

PR #65, opened 2018-05-10
Wire up adding/sharing a photo comment

asutula:


asutula:
This seems to be all working, but I'm not seeing the comments displayed after sharing.

andrewxhill:
okay, taking a look now

PR #72, opened 2018-05-11
small tweaks

andrewxhill:


App/Redux/TextileRedux.js
```diff
@@ -55,7 +55,7 @@ export const onboardedSuccess = state => {
 }
 
 export const handleImageAdded = (state, {thread, hash, remotePayloadPath}) => {
-  const items = [{ thread, hash, remotePayloadPath, state: 'pending' }, ...state.images.items]
+  const items = [...state.images.items, { thread, hash, remotePayloadPath, state: 'pending' }]
```
andrewxhill: figured we should just go with it the way we would and rn deal with it

App/Sagas/TextileSagas.js
```diff
@@ -209,7 +209,7 @@ export function * photosTask () {
   try {
     yield call(BackgroundTimer.start)
     const photos = yield call(queryPhotos)
-    for (const photo of photos) {
+    for (const photo of photos.reverse()) {
```
andrewxhill: this just allows your most recently taken photo to load last when bulk loading new images. it also makes the upload order appear a bit clearer

App/Navigation/PhotosNavigation.js
```diff
@@ -85,9 +85,9 @@ const TabNav = TabNavigator(
       const {routeName} = navigation.state
       let title
       if (routeName === 'PhotosNav') {
-        title = 'Library'
+        title = 'Wallet'
       } else if (routeName === 'SharedPhotosNav') {
-        title = 'Shared'
+        title = 'Threads'
```
andrewxhill: let me know what you think here

App/Redux/TextileRedux.js
asutula: huh? so you want oldest photos at the top? RN should/sounds-like-it-does behave fine either way, so we should just choose whichever we want.

App/Sagas/TextileSagas.js
asutula: Oh, then you switch it here

andrewxhill: unrelated. this is just the order within the new photos

asutula: or, this has nothing to do with display order, right?

App/Redux/TextileRedux.js
andrewxhill: no, what i saw was it doesn't matter actually. RN is putting our newest on top your way or my way

asutula: I don't see how that can be. Really?

asutula: Oh, this is the wrong file for this test. Sorry I didn't realize that earlier. Like i was saying in our chat, the authoritative source for the list of items is in `IpfsNodeRedux`. The list in here just serves as a pool of data to possibly augment items from the authoritative list with.

asutula: For the list in here, the order doesn't matter at all.

App/Sagas/TextileSagas.js
andrewxhill: so they one at a top pop into the displayed array. if you have your _newest_ pop in first, then others have to pop in behind it. that's pretty odd. here the oldest are going to pop in first. let me record you a gif

asutula: I get that, it's fine. This is a good change. This makes the photos pop in from oldest to newest. I like it.

asutula: Think we were miscommunicating  about what I meant by 'display order'. I meant the order of photos in the grid as a whole, while you were talking about the order in which they appear when added.

andrewxhill:
okay, gtg

PR #80, opened 2018-05-12
User avatars

andrewxhill:


asutula:
Last night I saw the hack to integrate jidenticon using those svg libs, was going to say it's not worth the trouble, but since you did it.... cool!

andrewxhill:
oh man, totally not worth the trouble. rn svg lib had linking problems and the works. but done now

App/Sagas/TextileSagas.js
```diff
@@ -162,11 +167,20 @@ export function * getPhotoHashes ({thread}) {
...
+        username = meta.username || 'anonymous'
         item = {...item, meta}
       } catch (err) {
         // gracefully return an empty meta for now
       }
-      data.push(item)
+      try {
+        const avatar = avatars[avatarHash] || jdenticon.toSvg(avatarHash, 200)
+        avatars[avatarHash] = avatar
+        item = {...item, avatar}
```
asutula: You see any good reason to store the svg avatar in the backing data when we can just derive it closer to the UI layer? I'm thinking the other place for it would be `mapStateToProps` in `TextilePhotos`.

andrewxhill: your call, i can move it

asutula: Not `TextilePhotos`. I meant `PhotoViewerScreen`. Or even in the render function of the metadata component in there.

andrewxhill: yeah sure, easy

App/Containers/PhotoViewerScreen.js
```diff
@@ -95,7 +124,7 @@ class PhotoViewerScreen extends React.PureComponent {
           onPageSelected={this.props.selectImage}
         />
         { this.galleryCount }
-        { this.caption }
+        { !this.props.sharable && this.caption }
```
andrewxhill: didn't make sense to show in own feed anymore

andrewxhill:
gtg

andrewxhill:
🤠

PR #81, opened 2018-05-12
add Events native module for go -> RN messages

sanderpick:
This uses a `Messenger` go interface that's implemented in Objective-C, which is passed to the node on creation. Because of this, the object itself cannot be a native module that implements `RCTEventEmitter`, since you're not able to directly allocate those objects. That is done by RN automatically. So, adopted an apparently popular pattern which uses a single `Events` native module which proxies to `NSNotificationCenter` (see SO link at top of `Events.h`). By doing this, we don't need the `TextileIPFS` module to extend the event emitter, since `Events` can handle all future events (you just call the static method `emitEventWithName`.

- general JS approach taken from here: https://blog.callstack.io/sending-events-to-javascript-from-your-native-module-in-react-native-29244f890e04
- I did not deal with actually calling the right JS method to update the thread, instead am just logging the object that details the update (not sure how to call the saga properly - found this with a quick search: https://github.com/redux-saga/redux-saga/issues/249)
- best docs on go side are just the go docs for `gobind`: https://godoc.org/golang.org/x/mobile/cmd/gobind

The 'could not decrypt file" error fix is in here too.


App/Containers/App.js
```diff
@@ -15,6 +16,13 @@ BackgroundTask.define(() => {
   store.dispatch(Actions.backgroundTask())
 })
 
+// subscribe to thread updates
+// TODO: this probably doesn't belong here
+// NOTE: we may want to cancel listener with the returned handle at some point with subscription.remove()
+IPFS.eventEmitter.addListener('onThreadUpdate', (payload) => {
+  console.log(payload)
```
sanderpick: again, not sure where this will live, placed in here for demo purposes.

TextileIPFSNativeModule.js
```diff
@@ -103,5 +108,10 @@ export default {
   getAccessToken: async function (): string {
     const result = await TextileIPFS.getAccessToken()
     return result
-  }
+  },
+
+  eventEmitter: Platform.select({
+    ios: new NativeEventEmitter(Events),
+    android: DeviceEventEmitter,
+  })
```
sanderpick: ios and android do this differently, this is taken from the article linked to in PR description

ios/NativeModules/TextileIPFS.m
```diff
@@ -12,14 +13,24 @@
...
-#endif
+@interface Messenger : NSObject<MobileMessenger>
+// Define class properties here with @property
+@end
+
+@interface Messenger()
+
+@end
+
+@implementation Messenger
```
sanderpick: here we go... send it to RN

ios/NativeModules/TextileIPFS.m
```diff
@@ -209,18 +220,15 @@ - (dispatch_queue_t)methodQueue {
...
 #pragma mark - Private methods
 
 - (void)_createNodeWithDataDir:(NSString *)dataDir apiUrl:(NSString *)apiUrl logLevel:(NSString *)logLevel error:(NSError**)error {
   if (!self.node) {
-    self.node = [[MobileMobile new] newNode:dataDir centralApiURL:apiUrl logLevel:logLevel error:error];
+    MobileNodeConfig *config = [[MobileNodeConfig alloc] init];
+    [config setRepoPath:dataDir];
+    [config setCentralApiURL:apiUrl];
+    [config setLogLevel:logLevel];
+    self.node = [[MobileMobile new] newNode:config messenger:[[Messenger alloc] init] error:error];
```
sanderpick: pass in our implementation of `MobileMessenger`

sanderpick:
Will need to get Java module in here too, but apparently that's easier. Will get to that this evening, unless somebody else wants to nip that.

andrewxhill:
trying this branch on iOS and it doesn't work for me unfortunately. images don't display, all blank. i tried `rm -rf node_modules` etc and no luck

android/app/src/main/java/com/textile/textileipfs/TextileIPFSModule.java
```diff
@@ -56,10 +59,25 @@ public String getName() {
...
+                NodeConfig config = new NodeConfig();
+                config.setRepoPath(dataDir);
+                config.setCentralApiURL(apiUrl);
+                config.setLogLevel(logLevel);
+                textile = Mobile.newNode(config, new Messenger() {
+                    @Override
+                    public void notify(Event event) {
+                        try {
+                            WritableMap payload = JsonConvert.jsonToReact(new JSONObject(event.getPayload()));
+                            TextileIPFSModule.emitDeviceEvent(event.getName(), payload);
```
sanderpick: emit event in Java

sanderpick:
okay guys, like i said, crucial concurrency fixes in here. going in.

sanderpick:
@asutula happy to revisit the bridges here, other than that this is pretty innocuous on the RN side

PR #86, opened 2018-05-14
holds onto async requests if node isnt started

andrewxhill:


andrewxhill:
well.. 2 things happen in this branch.

first: 1st is it unblocks the photogrid even if node isn't started.

as a result, you may do asyncimage requests prior to the node being available, so

second: it waits to send any asyncimage request until node is 'started'

andrewxhill:
biggest possible error that a user could get into now is that they make it all the way through a share before the node is started

App/Components/AsyncImage.js
```diff
@@ -53,3 +67,16 @@ export default class AsyncImage extends React.Component {
     }
   }
 }
+
+const mapStateToProps = (state, ownProps) => {
```
asutula: You don't have to make `AsyncImage` a redux connected component like this. You can derive this state>props for the node state up in `TextilePhotos`, then you just pass the `nodeState` as a prop into `PhotoGrid` and them again as a prop into `AsyncImage`.

andrewxhill:
Cool, lazy there. Done now.

andrewxhill:
gtg

PR #87, opened 2018-05-14
Fix Release Build Node Start

asutula:
Hack to detect when the app enters the foreground from cold start in a release build. I don't love this as a long term solution, but it should work for now. 

The key to the hack is that I found a particular pattern of behavior from `AppState` api when the app is launched into the foreground from a terminated state. We map this specific behavior to a 'foreground' event.

Let's be sure to test this on android as well, i've only tried iOS so far.

### DIFF
--- a/App/Containers/TextileManager.js
+++ b/App/Containers/TextileManager.js
@@ -32,6 +32,7 @@ class TextileManager extends React.PureComponent {
   }
 
   componentDidMount () {
+    this.props.appStateChange(this.props.currentAppState, AppState.currentState)
     BackgroundTask.schedule()
     // TODO: This logic should be moved deeper into the stack
     if (Platform.OS === 'android') {
--- a/App/Redux/IpfsNodeRedux.js
+++ b/App/Redux/IpfsNodeRedux.js
@@ -28,7 +28,7 @@ export default Creators
 
 export const INITIAL_STATE = Immutable({
   locked: false,
-  appState: 'unknown',
+  appState: null,
   nodeState: {
     state: 'undefined', // | creating | stopped | starting | started | stopping
     error: null
--- a/App/Sagas/StartupSagas.js
+++ b/App/Sagas/StartupSagas.js
@@ -35,6 +35,6 @@ export function * startup () {
 
   // Dispatch actions you want on STARTUP
   // yield put(IpfsNodeActions.createNodeRequest(RNFS.DocumentDirectoryPath))
-  const previousState = yield select(IpfsNodeSelectors.appState)
-  yield put(IpfsNodeActions.appStateChange(previousState, AppState.currentState))
+  // const previousState = yield select(IpfsNodeSelectors.appState)
+  // yield put(IpfsNodeActions.appStateChange(previousState, AppState.currentState))
 }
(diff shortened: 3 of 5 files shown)


App/Containers/TextileManager.js
```diff
@@ -32,6 +32,7 @@ class TextileManager extends React.PureComponent {
   }
 
   componentDidMount () {
+    this.props.appStateChange(this.props.currentAppState, AppState.currentState)
```
asutula: Was thinking that triggering the first state transition from a UI file would guarantee that `AppState.currentState` had a value by now. Not the case. It's still `unknown`, but let's use that to detect this specific case of launching from terminated into foreground.

App/Sagas/StartupSagas.js
```diff
@@ -35,6 +35,6 @@ export function * startup () {
 
   // Dispatch actions you want on STARTUP
   // yield put(IpfsNodeActions.createNodeRequest(RNFS.DocumentDirectoryPath))
-  const previousState = yield select(IpfsNodeSelectors.appState)
-  yield put(IpfsNodeActions.appStateChange(previousState, AppState.currentState))
```
asutula: Triggering this from `TextileManager` now.

App/Sagas/TextileSagas.js
```diff
@@ -86,15 +86,23 @@ export function * toggleBackgroundTimer ({value}) {
...
-    yield * triggerCreateNode()
-  } else if (previousState.match(/unknown|inactive|background/) && newState === 'active') {
-    console.tron.logImportant('app transitioned to foreground')
-    yield * triggerCreateNode()
-  } else if (previousState.match(/inactive|active/) && newState === 'background') {
-    console.tron.logImportant('app transitioned to background')
-    yield * triggerStopNode()
+  // TODO: HACK alert
+  if (!previousState) {
+    if (newState.match(/unknown|active/)) {
```
asutula: Here is the HACK. Gross. This `unknown` state is the signature of launching from scratch in a production build, it seems.

App/Sagas/index.js
```diff
@@ -63,7 +63,7 @@ export default function * root () {
     // takeEvery(StartupTypes.STARTUP, triggerCreateNode),
     // takeEvery(TextileTypes.LOCATION_UPDATE, triggerCreateNode),
     // takeEvery(TextileTypes.BACKGROUND_TASK, triggerCreateNode),
-    takeEvery(TextileTypes.ONBOARDED_SUCCESS, triggerCreateNode),
+    // takeEvery(TextileTypes.ONBOARDED_SUCCESS, triggerCreateNode),
```
asutula: Don't need this now that we're triggering the app state transition from `TextileManager` (which loads when onboarding is complete)

sanderpick:
Clever!

App/Redux/IpfsNodeRedux.js
```diff
@@ -28,7 +28,7 @@ export default Creators
 
 export const INITIAL_STATE = Immutable({
   locked: false,
-  appState: 'unknown',
+  appState: null,
```
carsonfarmer: this causes an error on android. i think the native module is expecting a string. should be able to catch it there i think

App/Sagas/StartupSagas.js
carsonfarmer: Is it still safe to have this here @asutula? Sticking this back in fixes the error on android.

PR #89, opened 2018-05-14
minor possible fix for login issue on android

carsonfarmer:
just undoes something that aaron added. i think its safe (its working on both os), but needs testing

### DIFF
--- a/App/Sagas/TextileSagas.js
+++ b/App/Sagas/TextileSagas.js
@@ -87,7 +87,6 @@ export function * toggleBackgroundTimer ({value}) {
 
 export function * handleNewAppState ({previousState, newState}) {
   // TODO: HACK alert
-  console.tron.logImportant(previousState, newState)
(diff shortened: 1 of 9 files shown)


App/Sagas/TextileSagas.js
```diff
@@ -102,6 +103,9 @@ export function * handleNewAppState ({previousState, newState}) {
     } else if (previousState.match(/inactive|active/) && newState === 'background') {
       console.tron.logImportant('app transitioned to background')
       yield * triggerStopNode()
+    } else if (previousState.match(/active/) && newState === 'active') {
```
carsonfarmer: so basically, we can call triggerCreateNode() almost be default any time we have a stateTransition, and just catch the one case where we are backgrounding right?

carsonfarmer: this one might not even be needed

PR #92, opened 2018-05-15
just a quick fix to make caption not look so bad

### LINKED ISSUE #91: Meta is coming back null in some cases
Observed after "violent backgrounding and foreground of the app"

screenshot 2018-05-14 10 56 00

andrewxhill:
okay, this fixes #91 for now, that whole flow will change so i didn't spend any time over thinking it

little bit more pop in avatar, rearranges things in places

screen shot 2018-05-14 at 6 33 54 pm


sanderpick:
oop, forgot to merge this

PR #102, opened 2018-05-17
Join the shared room on each node create -- build based room name and key

andrewxhill:


andrewxhill:
okay, this one is ready too

App/Sagas/TextileSagas.js
```diff
@@ -124,6 +125,7 @@ export function * createNode ({path}) {
     const debugLevel = (__DEV__ ? "DEBUG" : "INFO")
     const createNodeSuccess = yield call(IPFS.createNodeWithDataDir, path, API_URL, debugLevel)
     if (createNodeSuccess) {
+      yield call(IPFS.updateThread, Config.ALL_THREAD_MNEMONIC, Config.ALL_THREAD_NAME)
```
asutula: Should we be checking the return value here? If it fails, it should run the `else` below. So just to be safe, something like: move the call to `updateThread` out of this `if` and then update the `if` to `if (createNodeSuccess && updateThreadSuccess)`

ios/TextilePhotos.xcodeproj/xcshareddata/xcschemes/TextilePhotos.xcscheme
```diff
@@ -81,7 +90,7 @@
       </AdditionalOptions>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Debug"
+      buildConfiguration = "Release"
```
asutula: Did you change the scheme's 'run' action to `Release` and forget to change it back?

andrewxhill: so there are now two schemes. for the 'release' scheme, i guess you'd still want a debug build for logs etc?

asutula: Oh, right. Let me take a step back here and make sure I understand. Why do we need a new scheme? Within a single scheme, can't we load a different set of env vars based on whether it's a debug or release build? Maybe I don't understand the way react-native-config works...

asutula: I'm guessing that doesn't achieve some goal you had in mind, just want to understand what that is.

App/Sagas/TextileSagas.js
andrewxhill: cool, fixed

ios/TextilePhotos.xcodeproj/xcshareddata/xcschemes/TextilePhotos.xcscheme
asutula: I guess the way I'd think about the different schemes is not as 'release' and 'debug', but different environments, 'dev' and 'production'. You might want to run either scheme in debug or release mode depending on what you're doing. For that reason, I'd leave the 'run' action set to 'debug' so things behave as usual with Xcode development.

andrewxhill: cool, updated

PR #112, opened 2018-05-23
Removes SVG and Rect elements

### LINKED ISSUE #111: SVG icons cause crash on Android
Seems to also be documented here: https://github.com/facebook/react-native/issues/19113

Removing the SVG components from `PhotoViewerScreen` alleviates the problem. Simple solution would be to conditionally display the icon if on iOS...

carsonfarmer:
These were causing issues with rendering on Android. Thanks to @andrewxhill for discovering the fix. Since switching to a view instead of SVG, this should be consistent on both iOS and Android.

Fixes #111.


carsonfarmer:
Should be GTG @andrewxhill.

PR #116, opened 2018-05-24
[wip] an initial start to getting jest up and going

carsonfarmer:
starting with TextileIPFSNativeModule and moving on to actually testing the native module bridges. Then we should start adding components and snapshots etc (which will be part of separate PRs).


carsonfarmer:
Some testing resources:

- https://stackoverflow.com/questions/38003347/integration-testing-react-native-ios-unable-to-print-console-log
- https://blog.callstack.io/testing-your-react-native-apps-abfe41903dfd
- https://medium.com/@SkyscannerEng/testing-react-native-ios-bridges-80c730659a83

PR #121, opened 2018-05-25
Show image status + reset available retries after success

andrewxhill:
Couple things here. 

Tried some tests walking around the block with different connectivity and found it really hard to guess why a thumb or image would ever be missing. This branch now adds placeholders for different image statuses when they don't load. 'connecting' through 'error'. I think slow image loads mostly have to do with node unavailable yet, so this feedback is much nicer than just blank screen with no indication.

But another important thing I realized with the async image retries I pushed earlier today, a thumb could run out of retries even though it succeeded. Meaning if it later needs to refetch that data it may have run out of retries so just error out. This resets available retries if it succeeds.


App/Components/AsyncImage.js
```diff
@@ -22,59 +22,99 @@ export default class AsyncImage extends React.Component {
...
 
   _setSource = (source) => {
+    // After async token is received, it readys the image source
     if (!this.hasCanceled_) {
       this.setState(() => ({loaded: true, source}))
     }
   }
 
+  placeholder () {
+    if (this.state.error) {
```
asutula: Lots or repeated code in here. I'd just create a `let sourceFile` before the `if/else if` statements, the the value of `sourceFile` in the `if/if else` then just have the `<Image>` code once, and it uses `sourceFile`.

andrewxhill: nice

asutula:
Is this wip for a reason? Should I approve an merge?

andrewxhill:
i was still testing it out, just put out a TF to see how it looked in the wild. seems good to go. one thing it now highlights is the slow speed of full res over the gateway. it'll be good to implement the thumbnail preload then fullres to make that snappy.

andrewxhill:
gtg

PR #122, opened 2018-05-25
Update to background uploader with iOS improvements

asutula:


asutula:
Our build system does `yarn install` right?

andrewxhill:
yes

PR #124, opened 2018-05-26
Simple progressive load

andrewxhill:
Will always load thumb first and then full res only after it is successful


andrewxhill:
gtg

PR #129, opened 2018-05-29
Uri based cameraroll tracking

### LINKED ISSUE #127: New photos with old timestamps are ignored in upload
Using the timestamp as the getPhotos cursor allows downloaded/edited photos to sometimes mess up all uploading or not be uploaded themselves.

Needs a photo uuid based parsing of new photos

andrewxhill:
fixes #127

### DIFF
--- a/App/Redux/TextileRedux.js
+++ b/App/Redux/TextileRedux.js
@@ -9,7 +9,6 @@ const { Types, Creators } = createActions({
   locationUpdate: null,
   backgroundTask: null,
 
-  newImage: ['uri'],
   imageAdded: ['thread', 'hash', 'remotePayloadPath'],
 
   imageUploadProgress: ['data'],
@@ -42,7 +41,6 @@ export const INITIAL_STATE = Immutable({
   devices: []
 })
 
-
 /* ------------- Selectors ------------- */
 export const TextileSelectors = {
   // TODO: Add more selectors here as we learn how they are used
@@ -59,9 +57,11 @@ export const onboardedSuccess = state => {
   return state.merge({ onboarded: true })
 }
 
-export const handleImageAdded = (state, {thread, hash, remotePayloadPath}) => {
+export const handleImageAdded = (state, {uri, thread, hash, remotePayloadPath}) => {
+  const processed = state.camera && state.camera.processed ? state.camera.processed : []
+  const camera = {processed: [uri, ...processed]}
   const items = [{ thread, hash, remotePayloadPath, state: 'pending' }, ...state.images.items]
-  return state.merge({ images: { items } })
+  return state.merge({ images: { items }, camera })
 }
 
 export const handleImageProgress = (state, {data}) => {
@@ -143,12 +143,6 @@ export const pairNewDeviceError = (state, {pubKey}) => {
   return state.merge({ devices })
 }
 
-export const handleNewImage = (state, {uri}) => {
-  const processed = state.camera && state.camera.processed ? state.camera.processed : []
-  let camera = {processed: [uri, ...processed]}
-  return state.merge({ camera })
-}
-
 // Helper so sagas can figure out current items loaded
 // const getItems = state => state.items
 
@@ -157,7 +151,6 @@ export const handleNewImage = (state, {uri}) => {
 export const reducer = createReducer(INITIAL_STATE, {
   [Types.ONBOARDED_SUCCESS]: onboardedSuccess,
 
-  [Types.NEW_IMAGE]: handleNewImage,
   [Types.IMAGE_ADDED]: handleImageAdded,
 
   [Types.IMAGE_UPLOAD_PROGRESS]: handleImageProgress,
--- a/App/Redux/TextileRedux.js
+++ b/App/Redux/TextileRedux.js
@@ -9,14 +9,17 @@ const { Types, Creators } = createActions({
   locationUpdate: null,
   backgroundTask: null,
 
-  imageAdded: ['thread', 'hash', 'remotePayloadPath'],
+  imageIgnore: ['uri'],
+  imageAdded: ['uri', 'thread', 'hash', 'remotePayloadPath'],
 
   imageUploadProgress: ['data'],
   imageUploadComplete: ['data'],
   imageUploadCancelled: ['data'],
   imageUploadError: ['data'],
   imageRemovalComplete: ['id'],
+
   photosTaskError: ['error'],
+  photoProcessingError: ['id', 'error'],
 
   pairNewDevice: ['pubKey'],
   pairNewDeviceSuccess: ['pubKey'],
@@ -55,6 +58,13 @@ export const TextileSelectors = {
 
 export const onboardedSuccess = state => {
   return state.merge({ onboarded: true })
+}
+
+// Used to ignore certain URIs in the CameraRoll
+export const handleImageIgnore = (state, {uri}) => {
+  const processed = state.camera && state.camera.processed ? state.camera.processed : []
+  const camera = {processed: [uri, ...processed]}
+  return state.merge({ camera })
 }
 
 export const handleImageAdded = (state, {uri, thread, hash, remotePayloadPath}) => {
@@ -151,6 +161,7 @@ export const pairNewDeviceError = (state, {pubKey}) => {
 export const reducer = createReducer(INITIAL_STATE, {
   [Types.ONBOARDED_SUCCESS]: onboardedSuccess,
 
+  [Types.IMAGE_IGNORE]: handleImageIgnore,
   [Types.IMAGE_ADDED]: handleImageAdded,
 
   [Types.IMAGE_UPLOAD_PROGRESS]: handleImageProgress,
--- a/App/Redux/TextileRedux.js
+++ b/App/Redux/TextileRedux.js
@@ -68,7 +68,7 @@ export const onboardedSuccess = state => {
 // Used to ignore certain URIs in the CameraRoll
 export const handleImageIgnore = (state, {uri}) => {
   const processed = state.camera && state.camera.processed ? state.camera.processed : []
-  const camera = {processed: [uri, ...processed]}
+  const camera = {...state.camera, processed: [uri, ...processed]}
   return state.merge({ camera })
 }
 
@@ -77,7 +77,7 @@ export const toggleVerboseUi = state =>
 
 export const handleImageAdded = (state, {uri, thread, hash, remotePayloadPath}) => {
   const processed = state.camera && state.camera.processed ? state.camera.processed : []
-  const camera = {processed: [uri, ...processed]}
+  const camera = {...state.camera, processed: [uri, ...processed]}
   const items = [{ thread, hash, remotePayloadPath, state: 'pending' }, ...state.images.items]
   return state.merge({ images: { items }, camera })
 }
--- a/App/Redux/TextileRedux.js
+++ b/App/Redux/TextileRedux.js
@@ -145,7 +145,7 @@ export const pairNewDeviceError = (state, {pubKey}) => {
 
 export const handleNewImage = (state, {uri}) => {
   const processed = state.camera && state.camera.processed ? state.camera.processed : []
-  const camera = {processed: [uri, ...processed]}
+  let camera = {processed: [uri, ...processed]}
   return state.merge({ camera })
 }
 
--- a/App/Redux/TextileRedux.js
+++ b/App/Redux/TextileRedux.js
@@ -9,6 +9,7 @@ const { Types, Creators } = createActions({
   locationUpdate: null,
   backgroundTask: null,
 
+  newImage: ['uri'],
   imageAdded: ['thread', 'hash', 'remotePayloadPath'],
 
   imageUploadProgress: ['data'],
@@ -35,6 +36,9 @@ export const INITIAL_STATE = Immutable({
     loading: false,
     items: []
   },
+  camera: {
+    processed: []
+  },
   devices: []
 })
 
@@ -45,7 +49,8 @@ export const TextileSelectors = {
   itemsById: (state, id) => {
     return state.textile.images.items.filter(item => item.hash === id)
   },
-  onboarded: state => state.textile.onboarded
+  onboarded: state => state.textile.onboarded,
+  camera: state => state.textile.camera
 }
 
 /* ------------- Reducers ------------- */
@@ -138,6 +143,16 @@ export const pairNewDeviceError = (state, {pubKey}) => {
   return state.merge({ devices })
 }
 
+export const handleNewImage = (state, {uri}) => {
+  console.log('')
+  console.log(state.camera)
+  const processed = state.camera && state.camera.processed ? state.camera.processed : []
+  const camera = {processed: [uri, ...processed]}
+  console.log(camera)
+  console.log('')
+  return state.merge({ camera })
+}
+
 // Helper so sagas can figure out current items loaded
 // const getItems = state => state.items
 
@@ -146,6 +161,7 @@ export const pairNewDeviceError = (state, {pubKey}) => {
 export const reducer = createReducer(INITIAL_STATE, {
   [Types.ONBOARDED_SUCCESS]: onboardedSuccess,
 
+  [Types.NEW_IMAGE]: handleNewImage,
   [Types.IMAGE_ADDED]: handleImageAdded,
 
   [Types.IMAGE_UPLOAD_PROGRESS]: handleImageProgress,
--- a/App/Redux/TextileRedux.js
+++ b/App/Redux/TextileRedux.js
@@ -11,7 +11,7 @@ const { Types, Creators } = createActions({
   locationUpdate: null,
   backgroundTask: null,
 
-  imageIgnore: ['uri'],
+  urisToIgnore: ['uris'],
   imageAdded: ['uri', 'thread', 'hash', 'remotePayloadPath'],
 
   imageUploadProgress: ['data'],
@@ -43,9 +43,7 @@ export const INITIAL_STATE = Immutable({
     loading: false,
     items: []
   },
-  camera: {
-    processed: []
-  },
+  camera: {},
   devices: []
 })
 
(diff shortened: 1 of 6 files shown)


App/Services/PhotoUtils.js
```diff
@@ -1,96 +1,78 @@
...
+  return CameraRoll.getPhotos(queryParams)
+}
+
+export async function getAllPhotos () {
+  let data = []
+  let hasNextPage = true
+  let cursor = null
+  let totalPages = 50 // I wanted to limit the total size of a return array
+  while (hasNextPage && totalPages) {
+    let photos = await getPage(25, cursor)
```
andrewxhill: figured it would save us from having to do big array compares by just limiting this to 50 pages of 25. could likely go a lot smaller, but we can test it out

App/Services/PhotoUtils.js
```diff
@@ -1,96 +1,78 @@
...
+      // iOS method
+      if (photo.uri.includes('assets-library://')) {
+        var regex = /[?&]([^=#]+)=([^&#]*)/g, params = {}, match
+        while (match = regex.exec(photo.uri)) {
+          params[match[1]] = match[2]
         }
+        const path = fullDir + params.id + '.' + params.ext
+        await RNFS.copyAssetsFileIOS(photo.uri, path, 0, 0)
+        photo['path'] = path
+        const thumbPath = await resizeImage(photo.path, thumbRelativeDir)
```
andrewxhill: gotta move this guy. right now it'll always re-thumb existing files. coming up

App/Sagas/TextileSagas.js
```diff
@@ -239,9 +239,26 @@ export function * shareImage ({thread, hash, caption}) {
...
+    let allPhotos = yield call(getAllPhotos)
+    if (camera === undefined || camera.processed === undefined) {
+      for (const uri of allPhotos.map((p) => p.uri)) {
+        yield put(TextileActions.newImage(uri))
+      }
+      allPhotos = []
+    }
+
+    const newPhotos = allPhotos.filter((photo) => {
+      return processed.indexOf(photo.uri) === -1
```
asutula: I think we can optimize this big time. This is like O(n^2) since you have to iterate through `processed` for every `photo`. Check out https://github.com/textileio/textile-mobile/blob/master/App/Containers/TextilePhotos.js#L36. The basic idea is to convert the array of `processed` to an Object with a key for every item in `processed`. That way you can just do a direct lookup like `if (processedObject[photo.uri]) ...` and things are O(n).

App/Sagas/TextileSagas.js
```diff
@@ -239,9 +239,26 @@ export function * shareImage ({thread, hash, caption}) {
...
+    })
+
+    const photos = yield call(scalePhotos, newPhotos)
+
     for (const photo of photos.reverse()) {
       const multipartData = yield call(IPFS.addImageAtPath, photo.path, photo.thumbPath, 'default')
+      yield put(TextileActions.newImage(photo.uri))
```
asutula: Wondering if we need this new redux action `newImage` or can we just add a `uri` property to the `imageAdded` action that is triggered 3 lines below?

andrewxhill: okay, was putting it up here thinking through any errors we might hit below it wouldn't keep retrying. but happy to do that

asutula: Yea good call. Seems like a good idea, but that alone won't save off any error state for an individual photo because the `catch` we're in here is outside this `for` loop and just dispatches a general `photosTaskError`. To complete what you started here, there should be an inner (within the `for` loop) `try/catch` that catches errors for processing an individual photo, and dispatched a new action, `photoProcessingError(id, error)`, then the reducer can update the photo state with that error. 

If you want to add that here, cool, or we can save it for an additional PR.

andrewxhill:
okay, feeling good about this one but still would be good to test on a real device. can get that done first thing in the a.m.

andrewxhill:
okay, been playing around with this on device today. feels okay. i'd say gtg

App/Services/PhotoUtils.js
andrewxhill: ended up reducing this to just 250 photos max per camera roll update.

App/Redux/TextileRedux.js
```diff
@@ -54,9 +60,18 @@ export const onboardedSuccess = state => {
   return state.merge({ onboarded: true })
 }
 
-export const handleImageAdded = (state, {thread, hash, remotePayloadPath}) => {
+// Used to ignore certain URIs in the CameraRoll
+export const handleImageIgnore = (state, {uri}) => {
+  const processed = state.camera && state.camera.processed ? state.camera.processed : []
+  const camera = {processed: [uri, ...processed]}
```
asutula: Would be good to expand in the rest of the `camera` object in case we add anything to it in the future. If we leave this as as, any additional keys will be deleted. So something like:
```
const camera = {...state.camera, processed: [uri, ...processed]}
```

App/Redux/TextileRedux.js
```diff
@@ -54,9 +60,18 @@ export const onboardedSuccess = state => {
...
+  return state.merge({ camera })
+}
+
+export const handleImageAdded = (state, {uri, thread, hash, remotePayloadPath}) => {
+  const processed = state.camera && state.camera.processed ? state.camera.processed : []
+  const camera = {processed: [uri, ...processed]}
```
asutula: Same comment here as above.

App/Sagas/TextileSagas.js
```diff
@@ -239,24 +239,52 @@ export function * shareImage ({thread, hash, caption}) {
...
   try {
-    const photos = yield call(queryPhotos)
-    for (const photo of photos.reverse()) {
+    const camera = yield select(TextileSelectors.camera)
+    const processed = camera && camera.processed ? camera.processed : []
+
+    let allPhotos = yield call(getAllPhotos)
+    if (camera === undefined || camera.processed === undefined || processed.length === 0) {
+      const ignoredPhotos = allPhotos.splice(1)
+      for (const uri of ignoredPhotos.map((photo) => photo.uri)) {
```
asutula: No need to loop through all the photos twice, just loop through `ingnoredPhotos` and `yield put(TextileActions.imageIgnore(photo.uri))` (the map is the additional loop through)

App/Sagas/TextileSagas.js
```diff
@@ -239,24 +239,52 @@ export function * shareImage ({thread, hash, caption}) {
 
 export function * photosTask () {
   try {
-    const photos = yield call(queryPhotos)
-    for (const photo of photos.reverse()) {
+    const camera = yield select(TextileSelectors.camera)
+    const processed = camera && camera.processed ? camera.processed : []
+
+    let allPhotos = yield call(getAllPhotos)
+    if (camera === undefined || camera.processed === undefined || processed.length === 0) {
```
asutula: This can just be `if (processed.length === 0) {...}` because all the `undefined` checks are covered on line 243 above.

App/Sagas/TextileSagas.js
```diff
@@ -239,24 +239,52 @@ export function * shareImage ({thread, hash, caption}) {
...
+      const ignoredPhotos = allPhotos.splice(1)
+      for (const uri of ignoredPhotos.map((photo) => photo.uri)) {
+        yield put(TextileActions.imageIgnore(uri))
+      }
+    }
+
+    let allProcessed = processed.reduce((o, item, index) => ({...o, [item]: { index }}), {})
```
asutula: Nice! Storing an object containing the index as the value was just something needed for that example I pointed you to. The value could actually be anything. So it's fine as is, just an interesting point to mention...

App/Sagas/TextileSagas.js
```diff
@@ -239,24 +239,52 @@ export function * shareImage ({thread, hash, caption}) {
...
+    const photos = allPhotos.filter((photo) => {
+      if (allProcessed[photo.uri]) {
+        return false
+      }
+      return true
+    })
+
+    for (const orig of photos.reverse()) {
+      const photo = yield call(scalePhoto, orig)
```
asutula: This should be inside the `try`

App/Sagas/TextileSagas.js
```diff
@@ -239,24 +239,52 @@ export function * shareImage ({thread, hash, caption}) {
...
+
+    let allProcessed = processed.reduce((o, item, index) => ({...o, [item]: { index }}), {})
+    const photos = allPhotos.filter((photo) => {
+      if (allProcessed[photo.uri]) {
+        return false
+      }
+      return true
+    })
+
+    for (const orig of photos.reverse()) {
```
asutula: Just want to walk through what happens if the call to scalePhoto or unlink fails...

If I read this right, along with `TextileRedux`, if the call to scalePhoto or unlink fails a `photoProcessingError` is dispatched, but there is no reducer for that action, so no state is updated. But that's actually ok because the photo hasn't been added to the state yet. The next time the `photoTask` runs, the failed photos will be processed again.

App/Sagas/TextileSagas.js
```diff
@@ -239,24 +239,52 @@ export function * shareImage ({thread, hash, caption}) {
...
-      yield call(
-        Upload.startUpload,
-        {
-          customUploadId: multipartData.boundary,
-          path: multipartData.payloadPath,
-          url: 'https://ipfs.textile.io/api/v0/add?wrap-with-directory=true',
-          method: 'POST',
-          type: 'multipart',
-          field: multipartData.boundary
+      try { // single photo try/catch
```
asutula: Just want to walk through what happens if the call to scalePhoto or unlink fails...

If I read this right, along with TextileRedux, if the call to scalePhoto or unlink fails a photoProcessingError is dispatched, but there is no reducer for that action, so no state is updated. But that's actually ok because the photo hasn't been added to the state yet. The next time the photoTask runs, the failed photos will be processed again.

App/Sagas/TextileSagas.js
```diff
@@ -239,24 +239,52 @@ export function * shareImage ({thread, hash, caption}) {
...
+            type: 'multipart',
+            field: multipartData.boundary
+          }
+        )
+      } catch (error) {
+        try { // no matter, what. try to clean up photos
+          yield call(RNFS.unlink, photo.path)
+          yield call(RNFS.unlink, photo.thumbPath)
+        } finally {
+          yield put(TextileActions.photoProcessingError(multipartData.boundary, error))
```
asutula: Doesn't seem right that `photoProcessingError` is sometimes dispatched with a uri and sometimes with a hash.

asutula: Never mind, that's not true. Sorry.

App/Sagas/TextileSagas.js
```diff
@@ -239,24 +239,52 @@ export function * shareImage ({thread, hash, caption}) {
...
+    const photos = allPhotos.filter((photo) => {
+      if (allProcessed[photo.uri]) {
+        return false
+      }
+      return true
+    })
+
+    for (const orig of photos.reverse()) {
+      const photo = yield call(scalePhoto, orig)
       const multipartData = yield call(IPFS.addImageAtPath, photo.path, photo.thumbPath, 'default')
```
asutula: This inside the `try` as well

App/Sagas/TextileSagas.js
```diff
@@ -239,24 +239,52 @@ export function * shareImage ({thread, hash, caption}) {
...
+            path: multipartData.payloadPath,
+            url: 'https://ipfs.textile.io/api/v0/add?wrap-with-directory=true',
+            method: 'POST',
+            type: 'multipart',
+            field: multipartData.boundary
+          }
+        )
+      } catch (error) {
+        try { // no matter, what. try to clean up photos
+          yield call(RNFS.unlink, photo.path)
```
asutula: These calls to `unlink` seem redundant. Maybe the better structure for the singe photo try/catch would be, first remove the calls to `unlink` inside the `try`. Then the whole thing could look something like:
```
try {
  // scale, add, put imageAdded, startUpload
} catch {
  // put photoProcessingError
} finally {
  // unlink calls
}
```

App/Services/PhotoUtils.js
```diff
@@ -1,94 +1,72 @@
...
+    queryParams['after'] = cursor
+  }
+  return CameraRoll.getPhotos(queryParams)
+}
+
+export async function getAllPhotos () {
+  let data = []
+  let hasNextPage = true
+  let cursor = null
+  let totalPages = 10 // I wanted to limit the total size of a return array
```
asutula: Cool, any particular reason? Total edge case, but if you're on a river trip with no cell service and take more than 250 photos, we won't add all your new photos when you come back online.

App/Services/PhotoUtils.js
```diff
@@ -1,94 +1,72 @@
...
-  if (photosData.edges.length < 1) {
-    return
+
+  // iOS method
+  if (photo.uri.includes('assets-library://')) {
+    var regex = /[?&]([^=#]+)=([^&#]*)/g, params = {}, match
+    while (match = regex.exec(photo.uri)) {
+      params[match[1]] = match[2]
+    }
+    const path = fullDir + params.id + '.' + params.ext
```
asutula: Not sure if the merge will take care of this, but `params.ext` needs to be changed to `JPG`

App/Services/PhotoUtils.js
andrewxhill: yeah, was thinking through that. there is a performance hit as your existing images grow. to every foreground have to compare 1000s X 1000s arrays. so 250 felt like a comfortable number even for a rafting trip

App/Services/PhotoUtils.js
andrewxhill: probably out of the scope of the PR. i'm guessing we need some full logic there for jpg/gif/png yeah? and whatever android things to expect

App/Sagas/TextileSagas.js
andrewxhill: cool

App/Sagas/TextileSagas.js
andrewxhill: done

App/Redux/TextileRedux.js
andrewxhill: cool

andrewxhill:
above is gtg again

App/Services/PhotoUtils.js
asutula: I don't think it's so bad since you changed the code to just have to scan each array one time (one time to map the processed images to an object, and one time to check if each queried photo exists on that processed images object). That's linear performance O(n) and that's as good as it can get.

asutula: They way it was before was exponential performance overhead, which would have been a problem for sure.

App/Services/PhotoUtils.js
asutula: Well, this react-native-file-system always outputs a jpeg, so it's safe to hard code there I think.

App/Services/PhotoUtils.js
andrewxhill: upped it

App/Services/PhotoUtils.js
andrewxhill: done

App/Sagas/TextileSagas.js
```diff
@@ -239,25 +239,61 @@ export function * shareImage ({thread, hash, caption}) {
...
-          path: multipartData.payloadPath,
-          url: 'https://ipfs.textile.io/api/v0/add?wrap-with-directory=true',
-          method: 'POST',
-          type: 'multipart',
-          field: multipartData.boundary
+    const camera = yield select(TextileSelectors.camera)
+    const processed = camera && camera.processed ? camera.processed : []
+
+    let allPhotos = yield call(getAllPhotos)
+    if (processed.length === 0) {
```
asutula: Thought of one edge case we can totally ignore for now if you want, buuuut... 

The way this works, if the user has 0 photos in the camera roll, installs and runs Textile, takes some photos, opens Textile, those photos will be added to the ignored list, and only after that will subsequent photos be processed.

asutula: The fix would be for `camera.processed === null` to represent the state where the initial 'ignore photos' query has not yet been run. Once `camera.processed` exists, even if it is empty, there would be no need to query photos to ignore.

App/Sagas/TextileSagas.js
```diff
@@ -239,25 +239,61 @@ export function * shareImage ({thread, hash, caption}) {
...
+      if (allProcessed[photo.uri]) {
+        return false
+      }
+      return true
+    })
+
+    for (const orig of photos.reverse()) {
+      let photo
+      try { // single photo try/catch
+        photo = yield call(scalePhoto, orig)
```
asutula: This can go in the `try` below with everything else, right?

App/Sagas/TextileSagas.js
```diff
@@ -239,25 +239,61 @@ export function * shareImage ({thread, hash, caption}) {
...
+      } catch (error) {
+        yield put(TextileActions.photoProcessingError(orig.uri, error))
+        continue // jump to next photo
+      }
+
+      // ** Only runs if photo scaling was successful **
```
asutula: If you move the call to scale the photo into this `try`, the desired behavior described here will be true... If the photo scaling fails, all the subsequent code inside the `try` won't run, it will just call `catch` and `finally` and then control will go back to the `for` loop and the next iteration of the loop will run. I think that's right.

andrewxhill: the reason i did it this way is so that i was safe to call `call(RNFS.unlink, photo.path)` in the finally below. if i put this in the same try and it failed the call to finall unlink would also fail

andrewxhill: or i move the scale and then wrap the finally content in a new try, you see?

App/Sagas/TextileSagas.js
andrewxhill: okay, this edge case has occurred but thought so little hadn't thought of this solution. will try quickly here

App/Sagas/TextileSagas.js
asutula: You already have the `unlink` in the `finally` wrapped in another try/catch, right?

andrewxhill: ah, the looking closer, the real reason was thinking through what information in the error would actually be actionable. i can just change all errors to deal with uri as the identifier and then  yours works

andrewxhill: coming up

asutula: cool

App/Sagas/TextileSagas.js
```diff
@@ -239,25 +239,53 @@ export function * shareImage ({thread, hash, caption}) {
...
+        return false
+      }
+      return true
+    })
+
+    for (let photo of photos.reverse()) {
+      try { // single photo add and upload
+        photo = yield call(scalePhoto, photo)
+        const multipartData = yield call(IPFS.addImageAtPath, photo.path, photo.thumbPath, 'default')
+        yield put(TextileActions.imageAdded(photo.uri, 'default', multipartData.boundary, multipartData.payloadPath))
```
andrewxhill: You had originally asked that I join the photo.uri to be added to camera.processed in this single call @asutula, but that means it will be retried. sounds good but probably needs more thought when to retry. my gut says we actually want to add it to the processed in the event of errors, so in the finally, so that we don't retry it forever. if you think like this is better, all good, otherwise i can mod it quickly

andrewxhill: like this === your way, just add it when we do the imageAdded and retry on any errors

asutula: Yea, so now if any error happens when processing the photo, we just 'forget' about the fact that we ever queried the photo and tried to process it. The next time `photosTask` runs, the failed photo will be fetched and processed again. Right?

I think that's fine for now. We can put more thought into a smarter system for retrying later on. It's not simple.

andrewxhill: yeah, that's right. cool

PR #131, opened 2018-05-30
new frameworks with small change for pinging threads

sanderpick:
goes with https://github.com/textileio/textile-go/pull/129


sanderpick:
textileio/textile-go#129 is merged

PR #132, opened 2018-05-30
Node status viz

asutula:
This adds a 'preferences' key to our persistent Redux. Right now the only key in there is `verboseUi`. When `verboseUi` is `true`, we can chose to render more data into the UI. For now, I display the node state at the bottom of the photo grid so you can see it even when there are photos in the grid.


asutula:
Oh, important detail... The way you toggle `verboseUi` on and off is by long pressing on the 'Textile' logo at the top of the main screen for three seconds.

PR #135, opened 2018-06-01
Preferences persistence migration

asutula:
Turns out the behavior of redux-persist is to overwrite your initial state (and any changes/additions you make to it in, for example, `TextileRedux.js`) with whatever was persisted previously. This explains why the `textile.preferences.verboseUi` key wasn't present when updating to the version of the app that added it to `TextileRedux.js`... The previously persisted data didn't contain that key.

The answer: Migration.

This looks like a tiny change, but I dug deep into understanding how redux, redux-persist, seamless-immutable and migrations all work together. Pretty tricky stuff, but now that I understand it, we're gold for migrating persisted data in the future.

### DIFF
--- a/App/Config/ReduxPersist.js
+++ b/App/Config/ReduxPersist.js
@@ -4,7 +4,7 @@ import { createMigrate } from 'redux-persist'
 
 const migrations = {
   0: (state) => {
-    // migration clear out device state
+    // Migration to add user preferences with option for verboseUi
     return {
       ...state,
       textile: state.textile.merge({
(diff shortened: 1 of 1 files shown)


App/Config/ReduxPersist.js
```diff
@@ -1,13 +1,30 @@
...
+  }
+}
 
 // More info here:  https://shift.infinite.red/shipping-persistant-reducers-7341691232b1
 const REDUX_PERSIST = {
   active: true,
   reducerVersion: '1.0',
   storeConfig: {
     key: 'primary',
+    version: 0,
```
andrewxhill: so simple in the end. nice

PR #136, opened 2018-06-01
Fix node stopped state

sanderpick:
I had @andrewxhill sneak the sagas change into the App Store release because, without it, my app was not usable (always stopped after a foreground). Tested on debug and release builds.

Also, this allows setting a `logFile` option from RN. Coupled with the new node state UI, this makes for a super useful way to understand what's actually happening when running on device. This wasn't possible before because they only way to get at the textile log file when running on device was to download the entire container (at least on iOS) and try to match dates when you did things... no good.

### DIFF
--- a/App/Sagas/TextileSagas.js
+++ b/App/Sagas/TextileSagas.js
@@ -114,11 +114,6 @@ export function * handleNewAppState ({previousState, newState}) {
 }
 
 export function * triggerCreateNode () {
-  const locked = yield select(IpfsNodeSelectors.locked)
-  if (locked) {
-    return
-  }
-  yield put(IpfsNodeActions.lock(true))
   yield put(IpfsNodeActions.createNodeRequest(RNFS.DocumentDirectoryPath))
 }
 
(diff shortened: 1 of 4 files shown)


App/Sagas/TextileSagas.js
```diff
@@ -114,11 +114,6 @@ export function * handleNewAppState ({previousState, newState}) {
 }
 
 export function * triggerCreateNode () {
-  const locked = yield select(IpfsNodeSelectors.locked)
```
asutula: Seems like if we're going to remove this use of `lock`, we should remove it everywhere.

App/Sagas/TextileSagas.js
```diff
@@ -128,8 +123,9 @@ export function * triggerStopNode () {
 
 export function * createNode ({path}) {
   try {
-    const debugLevel = (__DEV__ ? 'DEBUG' : 'INFO')
-    const createNodeSuccess = yield call(IPFS.createNodeWithDataDir, path, API_URL, debugLevel)
+    const logLevel = (__DEV__ ? 'DEBUG' : 'INFO')
+    const logFiles = !__DEV__
```
asutula: Does `logFiles === true` mean that the log output shows up in Xcode or not? Kind of a strange name.

sanderpick: covered in our call, but adding here for future ref... `logFiles` is a boolean indicating whether or not to write log output to actual files on the file system.

sanderpick: (if false, they will instead be written to stdout)

PR #139, opened 2018-06-04
Progressive load in shared thread

andrewxhill:


App/Containers/PhotoViewerScreen.js
```diff
@@ -33,7 +33,7 @@ class PhotoViewerScreen extends React.PureComponent {
     return (<AsyncImage
       key={props.image.key}
       hash={props.image.hash}
-      progressiveLoad={props.image.progressiveLoad}
+      progressiveLoad='true'
```
asutula: This is probably working for the wrong reason inside `AsyncImage` (would have the same result if set to `'false'`. Should be `progressiveLoad={true}`.

andrewxhill: ah, misread what you were saying there. fixed that up. but still hold off on rest for a minute

andrewxhill:
Okay, needed an overhaul. Here's what I found,

Progressive load on your own thread worked fine because if Thumbs worked, Full res always worked too. So, after the thumb loaded I would just swap the src in for the full res and it would load up.

But, doing that on the shared Thread was a bit tricky. It meant that if IPFS couldn't find or download the full res in time, the user would be stuck looking at blank screen since the src was swapped already. And then... falling back to the thumb would require a whole new request. blah blah annoying.

Took a look at a few other examples and found that stacking a full res over the top of the thumbnail was a common way to do it. If the full res failed, just keep it transparent and the world is fine. I implemented that here, took some reconfiguring. 

I've split the logic between `IPFSImage` which knows how to stack two `AsyncImages` and how to call the right stack for a single hash.

sanderpick:
How's this looking to you @asutula?

PR #142, opened 2018-06-05
Allows images to preprocess and then updates gallery 1 time

andrewxhill:
Changes the update loop to only update the gallery 1 time once all the photos are pre-processed

### DIFF
--- a/App/Sagas/TextileSagas.js
+++ b/App/Sagas/TextileSagas.js
@@ -264,12 +264,36 @@ export function * photosTask () {
       return true
     })
 
+
+    let photoUploads = []
+    // Convert all our new entries to thumbs before anything else
     for (let photo of photos.reverse()) {
-      try { // single photo add and upload
+      try {
         photo = yield call(scalePhoto, photo)
         const multipartData = yield call(IPFS.addImageAtPath, photo.path, photo.thumbPath, 'default')
+        photoUploads.push({
+          photo: photo,
+          multipartData
+        })
+      } catch (error) {
+        // if error, delete the photo copy and thumb from disk then fire error
+        try {
+          yield call(RNFS.unlink, photo.path)
+          yield call(RNFS.unlink, photo.thumbPath)
+        } finally {
+          yield put(TextileActions.photoProcessingError(photo.uri, error))
+        }
+      }
+    }
+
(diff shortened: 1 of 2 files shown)


App/Sagas/TextileSagas.js
```diff
@@ -264,12 +264,36 @@ export function * photosTask () {
...
+          yield call(RNFS.unlink, photo.path)
+          yield call(RNFS.unlink, photo.thumbPath)
+        } finally {
+          yield put(TextileActions.photoProcessingError(photo.uri, error))
+        }
+      }
+    }
+
+    // refresh our gallery
+    yield put(IpfsNodeActions.getPhotoHashesRequest('default'))
```
asutula: Wondering why this is up here instead of after the loop below (`imageAdded` is dispatched in the loop below). Think it will be more efficient (fewer renders) to put it below that loop.

andrewxhill: how much efficiency do you gain there though? the trade of is that with this setup the user sees their updates faster upon opening the app, which is what i was going for

asutula:
Hmm, not sure what i think about the UX of this. Imagine you need to process lots of image, extreme case being where we have an option up load all old photos when you first install... You'll be staring at a blank screen for a long time then suddenly all your photos will appear. What's wrong with having them appear as they are processed?

App/Sagas/TextileSagas.js
asutula: I guess i don't understand how this works. Seems like it would be slower to me since you're delaying updating the redux state by processing all added images before calling `imageAdded`.

asutula: If you want the fastest images on the screen, you should move `getPhotoHashesRequest` to the loop above right after each `addImageAtPath` is called.

asutula: Sorry i thought this PR was more about UI performance, not UX. Yea for the fastest UX do what I said above.

asutula: Hmm actually it seems like getPhotoHashesRequest runs slow (?) sometimes. So maybe leaving it where it is is fine. Thanks for letting me thing through this out loud.

PR #143, opened 2018-06-06
IpfsImage Component

asutula:
Idea here is to create a view component that behaves exactly like the React Native `Image`, but instead of supplying a URI or base64 string as the image source, you specify an IPFS hash path.

Since the API is totally RN, ObjC/Java based, there are no security issues (like when we had to run a http gateway to serve the images), no request token, auth, etc.

I basically copied all the `Image` related code from RN and modified it to use our IPFS native module. This guarantees that we're using the same instance of the native module from this new `IpfsImage` component as well as other places in the app (adding image, for example)... There is only one native module instance created by RN, and therefore only IPFS node instance created.


App/Components/PhotoGrid.js
```diff
@@ -42,12 +43,13 @@ export default class PhotoGrid extends React.PureComponent {
...
-            resizeMode={'cover'}
-            displayImages={this.props.displayImages}
-          />
+          {/*<IPFSImage*/}
+            {/*hash={row.item.hash}*/}
+            {/*style={styles.itemImage}*/}
+            {/*resizeMode={'cover'}*/}
+            {/*displayImages={this.props.displayImages}*/}
+          {/*/>*/}
+          <IpfsImage source={{hashPath: row.item.hash + '/thumb'}} style={styles.itemImage} resizeMode={'cover'} />
```
asutula: Any props you can pass to a normal `Image` should continue working.

ios/NativeModules/TextileIPFS.h
```diff
@@ -9,6 +9,17 @@
...
 @interface TextileIPFS : NSObject <RCTBridgeModule>
   // Define class properties here with @property
+- (NSString *)_getHashData:(NSString *)hashPath error:(NSError**)error;
+@end
+
+@interface RCTBridge (TextileIPFS)
+/**
+ * The shared ipfs instance
+ */
+@property (nonatomic, readonly) TextileIPFS *ipfs;
```
asutula: Here we extend the RN bridge object and add a property to retrieve the single instance of the Ipfs natvie module. The bridge is a shared object that's accessible from all other native modules and native UI components, therefore this allows us to access the IPFS node API from any native module or native UI component we create.

ios/NativeModules/TextileIPFS.m
```diff
@@ -309,3 +309,12 @@ - (NSString *)_getAccessToken:(NSError**)error {
 }
 
 @end
+
+@implementation RCTBridge (TextileIPFS)
+
+- (TextileIPFS *)ipfs
+{
+  return [self moduleForClass:[TextileIPFS class]];
```
asutula: This is how you return the single instance of the IPFS node API (our `TextileIPFS` native module). `moduleForClass` is basically a service lookup that returns the single instance of any native module based on class type.

ios/NativeViews/IpfsImageSource.h
```diff
@@ -0,0 +1,35 @@
+#import <Foundation/Foundation.h>
+
+#import <React/RCTConvert.h>
+
+/**
+ * Object containing an image hash path and associated metadata.
+ */
+@interface IpfsImageSource : NSObject
```
asutula: This is the class that models the backing data for `IpfsImage`. Looks just like `RCTImage` source, but with the `NSURLRequest` swapped out for a `NSString` for the hash path.

IpfsImage.js
```diff
@@ -0,0 +1,267 @@
...
+const IpfsImageViewManager = NativeModules.IpfsImageViewManager;
+
+/**
+ * A React component for displaying different types of images,
+ * including network images, static resources, temporary local images, and
+ * images from local disk, such as the camera roll.
+ *
+ * See https://facebook.github.io/react-native/docs/image.html
+ */
+const IpfsImage = createReactClass({
```
asutula: This is the JS layer of the new `IpfsImage`. This is actually the iOS version, still need to make the android version (RN uses different JS implementations of `Image` depending on platform).

sanderpick:
Super duper. What do you think about calling it `TextileImage`? Only reason is that folks might try to use this for normal images hosted on IPFS, whereas this does decryption. Similarity, the "node" isn't really an IPFS node... of course it contains one, but the distinction is significant and growing.

ios/NativeViews/IpfsImageView.h
```diff
@@ -0,0 +1,19 @@
+#import <UIKit/UIKit.h>
+
+#import <React/RCTResizeMode.h>
+
+@class RCTBridge;
+@class IpfsImageSource;
+
+@interface IpfsImageView : UIImageView
+
+- (instancetype)initWithBridge:(RCTBridge *)bridge NS_DESIGNATED_INITIALIZER;
```
asutula: This is the Obj C native class that is exported to RN and backs `IpfsImage` JS class.

sanderpick:
We'll have to revisit slightly when the v2 stuff is in because the hash alone won't be enough... you'll need to pass something like `<thread_pk_base64>:<content_hash>/<photo|thumb|etc>`

ios/NativeViews/IpfsImageView.m
```diff
@@ -0,0 +1,460 @@
...
+//                                           scale:imageScale
+//                                         clipped:NO
+//                                      resizeMode:_resizeMode
+//                                   progressBlock:progressHandler
+//                                partialLoadBlock:partialLoadHandler
+//                                 completionBlock:completionHandler];
+
+    NSError *error;
+    UIImage *image;
+    NSString *base64String = [_bridge.ipfs _getHashData:source.hashPath error:&error];
```
asutula: Even though our API for loading image data is synchronous, I left all the async handling from the `URLRequest`-based `Image` code just in case we switch our API to async some day. Seems like that might be a good idea.

ios/NativeViews/IpfsImageViewManager.h
```diff
@@ -0,0 +1,5 @@
+#import <React/RCTViewManager.h>
+
+@interface IpfsImageViewManager : RCTViewManager
```
asutula: This manager class is just standard RN native UI component stuff... You have to create the actual view implementation plus a 'manager' class that provides the RN runtime with the metadata it needs to be able to register and use your component.

asutula:
For reference, here is a list of all the classes I had to understand/copy/modify in order to get iOS working. We'll have to do the same for Android:
```
TextileIPFS (to add the RCTBridge extension to expose TextileIPFS)
RCTConvert (to add support converting to/from IpfsImageSource)
RCTImageSource
RCTImageView
RCTImageViewManager
Image.ios.js (RN's iOS JS API)
Image.android.js (RN's Android JS API)
```

asutula:
Cool @sanderpick, yea calling it `TextileImage` is a good idea. I'll do that now.

Adding params to the image source is no problem at all.

sanderpick:
Awesome

PR #144, opened 2018-06-08
Progressive loading of TextileImage

andrewxhill:
…plays thumb until full res is loaded


ios/NativeViews/TextileImageView.m
```diff
@@ -318,16 +318,18 @@ - (void)reloadImage
...
-    UIImage *image;
-    NSString *base64String = [_bridge.textileNode _getHashData:source.hashPath error:&error];
-    if (base64String) {
-      NSString *finalBase64String = [@"data:image/jpeg;base64," stringByAppendingString:base64String];
-      NSURL *url = [NSURL URLWithString:finalBase64String];
-      NSData *imageData = [NSData dataWithContentsOfURL:url];
-      image = [UIImage imageWithData:imageData];
-    }
-    completionHandler(error, image);
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
andrewxhill: Added this to allow the image data request to enter another thread. https://github.com/facebook/react-native/issues/14841

App/Components/ProgressiveImage.js
```diff
@@ -0,0 +1,80 @@
...
+      return (
+        <View style={this.props.style}>
+          {this.renderPhoto()}
+        </View>
+      )
+    } else if (this.state.preview) {
+      // if thumb is available, start loading the full res
+      return (
+        <View style={this.props.style}>
+          {this.renderPreview}
```
asutula: Reads odd to have `renderPreview` as a `get` function and `renderPhoto` as a normal function. I'd just make the both normal functions.

App/Sagas/TextileSagas.js
```diff
@@ -264,12 +264,36 @@ export function * photosTask () {
       return true
     })
 
+
+    let photoUploads = []
```
asutula: This change looks familiar... Did this all just come from merging in our previous work around triggering a single UI update for all added photos?

ios/NativeViews/TextileImageView.m
asutula: Yea, cool. This makes sense. Now that we're calling these methods directly from UIKit code (our native image view), the method is being called on the main thread and blocking the UI. Previously, any method in here was just called from the JS layer, and not on the main thread.

App/Sagas/TextileSagas.js
andrewxhill: The target of this MR is your WIP, maybe it wasn't up to date with master yet?

andrewxhill:
removed the 'get' method

ios/NativeViews/TextileImageView.m
asutula: Oh, just realized this is actually in the native image class... Cool, this works!

PR #147, opened 2018-06-10
Photo processing with state

### LINKED ISSUE #146: Partial processing of camera roll leads to occasional duplicate entries.
See this a lot when running app in background (maybe offline) mode while taking a lot of photos. The final images don't seem to load very successfully and often have duplicates.

andrewxhill:
fixes #146 

this feels a lot better

### DIFF
--- a/App/Config/ReduxPersist.js
+++ b/App/Config/ReduxPersist.js
@@ -25,6 +25,19 @@ const migrations = {
         images: state.textile.images.merge({ items: updatedItems })
       })
     }
+  },
+  2: (state) => {
+    const uris = state.textile.camera && state.textile.camera.processed ? state.textile.camera.processed : []
+    let processed = {}
+    for (let uri of uris) {
+      processed[uri] = 'complete'
+    }
+    return {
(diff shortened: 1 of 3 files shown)


asutula:
If this isn't urgent, I'd ask to hold off on this and let me get around to the general cleanup of the photo query trigger and photo processing code I started a while back and need to wrap up. This looks like it introduces some new data model and migrations, than may not be needed in my WIP on this stuff and could just complicate that WIP. Totally open to discussing this further if there's some urgent fix here that I need to understand better.

andrewxhill:
My app was broken over the weekend, stuck trying to reprocess the same images (and slowly exploding my storage). figuring that a lot of people do what i did on the weekend I pushed this fix to the app. So we'll need a migration either way, sorry about that :(

asutula:
Like this is already out in the App Store even though it's still on a branch in an PR?

andrewxhill:
yeah, it was a real problem and you were out (wasn't going to merge your branch into master)

PR #155, opened 2018-06-20
Integrate Textile Node V2

### LINKED ISSUE #152: Update mobile bridges for v2 arch

sanderpick:
fixes #152


sanderpick:
Put back in WIP... couple small TODOs:
- [ ] - brute migration (delete db)
- [ ] - fix that all thread thingy nit (just needs a diff name in prod, “all_prod” or whatever)
- [ ] - make a global update room so we can try ditching the relays in a really crude manner (can come after we merge this, just at the lib layer)

sanderpick:
okay @asutula - this one is cooked. included frameworks from https://github.com/textileio/textile-go/pull/147. sharing will currently not work too well since the relays need an update, but other than that would be good to start running this on the team. I can update the relays tonight / tomorrow.

asutula:
Awesome, will take a look this eve.

App/Containers/PhotoViewerScreen.js
```diff
@@ -186,7 +187,7 @@ const mapStateToProps = (state, ownProps) => {
 
 const mapDispatchToProps = (dispatch) => {
   return {
-    share: (item) => { dispatch(IpfsActions.shareImageRequest('all', item.hash)) },
+    share: (item) => { dispatch(IpfsActions.shareImageRequest(Config.ALL_THREAD_NAME, item.hash)) },
```
asutula: Didn't we decide this `IpfsActions.shareImageRequest` doesn't exist anymore?

sanderpick: oh yeah. what can be deleted here? `share`, `authorShare`, `cancelAuthoringShare`?

App/Sagas/TextileSagas.js
```diff
@@ -218,20 +220,8 @@ export function * pairNewDevice (action) {
 
 export function * shareImage ({thread, hash, caption}) {
   try {
-    const multipartData = yield call(IPFS.sharePhoto, hash, thread, caption)
-    yield put(TextileActions.imageAdded(thread, multipartData.boundary, multipartData.payloadPath))
+    yield call(IPFS.sharePhoto, hash, thread, caption)
     yield put(IpfsNodeActions.getPhotoHashesRequest(thread))
-    yield call(
```
asutula: Awesome to get rid of this

.env
```diff
@@ -1,4 +1,4 @@
 # required for react-native-config
 TEXTILE_API_URI="https://api.textile.io"
 ALL_THREAD_MNEMONIC="track soccer chapter great stove copy forum donate immune cattle boost action menu buyer mention spoon bacon boss suffer awful outdoor drum picture glance"
-ALL_THREAD_NAME="all"
+ALL_THREAD_NAME="all_development"
```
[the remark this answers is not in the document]
sanderpick: prod will be the same, but what's the value of that variable in CI? looked, but not able to see the value beyond "xxxxIgo=". Is it just the all thread name or mnemonic?

App/Sagas/TextileSagas.js
```diff
@@ -265,8 +255,8 @@ export function * photosTask () {
     // Convert all our new entries to thumbs before anything else
     for (let photo of photos.reverse()) {
       try {
-        photo = yield call(scalePhoto, photo)
-        const multipartData = yield call(IPFS.addImageAtPath, photo.path, photo.thumbPath, 'default')
+        photo = yield call(getPhotoPath, photo)
+        const multipartData = yield call(IPFS.addPhoto, photo.path, 'default', '')
```
asutula: Always suspicious about using empty strings... why not `null`?

sanderpick: agreed yeah... in this case though, it's just the caption. an "add" from the mobile perspective is really 1) add the file to IPFS, 2) essentially "share" it to a thread (in this case default). I left the caption arg here because at some point you may not want to _first_ add a photo to your default thread, maybe you want to just send it direct to some other thread. at some point the go layer would have to convert `nil` to "", encrypt that and add it to the block as a file

App/Containers/PhotoViewerScreen.js
asutula: Looks like `share` and `cancelAuthoringShare` can be removed here.

App/Sagas/TextileSagas.js
sanderpick: yep, fair enough, can change it np

.env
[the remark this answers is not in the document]
asutula: As in, the production values aren't being updated at all?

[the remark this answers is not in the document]
sanderpick: yep, in prod, the all thread name is still "all" and the mnemonic is unchanged.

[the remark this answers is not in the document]
asutula: cool

PR #156, opened 2018-06-21
Fix Android dev mode compile

asutula:
If keystore env var isn’t set, just provide a string so dev compile can work


asutula:
Gonna merge this, simple, need it to keep working now

PR #157, opened 2018-06-22
Litte bug fix to resolve promise with true

asutula:


asutula:
Been meaning to clean up the underlying promises and the code that calls them. Currently, the calling code awaits the boolean value fulfilled by the promise, so passing `null` was causing the a failure condition. Buuuuuut... Maybe the long term fix is to make them fulfill with `Void` or whatever and just know the error can be caught.

sanderpick:
i like that sound of the latter solution... just looks weird returning with true... but understandable, it's a weird collision of worlds

PR #160, opened 2018-06-22
Node fixes

asutula:
Just cleaning up how we interact with the JS node API, fixing a couple bugs. Started down the path of adding tests for the native layer node modules, but it's harder than I thought, will tackle it in another PR.


asutula:
Ok I think this is g2g. Will tackle native module testing in a new PR.

App/Sagas/TextileSagas.js
```diff
@@ -220,6 +203,7 @@ export function * pairNewDevice (action) {
 
 export function * shareImage ({thread, hash, caption}) {
   try {
+    // TODO: sharePhoto returns a String. Should we be using it?
```
sanderpick: the return string here is the newly created block id, we can remove that at some point because it doesn't seem useful since you just request all photos again

android/app/src/main/java/com/textile/textilenode/TextileNode.java
```diff
@@ -272,8 +277,7 @@ public void getFileData (String id, String path, Promise promise) {
     @ReactMethod
     public void pairDevice (String pkb64, Promise promise) {
         try {
-            node.pairDevice(pkb64);
-            promise.resolve(true);
+            promise.resolve(node.pairDevice(pkb64));
```
sanderpick: awesome

PR #161, opened 2018-06-26
Android Node Query Fixes

asutula:
Android got a little out of whack with the last node update. This just fixed the problems.

### DIFF
--- a/android/app/src/main/java/com/textile/textilenode/TextileNode.java
+++ b/android/app/src/main/java/com/textile/textilenode/TextileNode.java
@@ -242,10 +242,13 @@ public class TextileNode extends ReactContextBaseJavaModule {
             // convert response to json
             JSONObject obj = new JSONObject(jsonString);
             WritableArray array = new WritableNativeArray();
-            JSONArray jsonArray = obj.getJSONArray("items");
+            JSONArray jsonArray = obj.optJSONArray("items");
+            if (jsonArray == null) {
+                jsonArray = new JSONArray();
+            }
             for (int i = 0; i < jsonArray.length(); i++) {
                 Object value = jsonArray.get(i);
-                array.pushString((String) value);
+                array.pushString(value.toString());
             }
             promise.resolve(array);
         }
--- a/android/app/src/main/java/com/textile/textilenode/TextileNode.java
+++ b/android/app/src/main/java/com/textile/textilenode/TextileNode.java
@@ -9,23 +9,17 @@ import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactContextBaseJavaModule;
 import com.facebook.react.bridge.ReactMethod;
-import com.facebook.react.bridge.ReadableArray;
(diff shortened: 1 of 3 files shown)


android/app/src/main/java/com/textile/textilenode/TextileNode.java
```diff
@@ -241,12 +246,11 @@ public void getPhotoBlocks (String offset, Integer limit, String threadName, Pro
...
-            JSONArray jsonArray = obj.getJSONArray("items");
-            for (int i = 0; i < jsonArray.length(); i++) {
-                Object value = jsonArray.get(i);
-                array.pushString((String) value);
+            JSONArray jsonArray = obj.optJSONArray("items");
+            if (jsonArray == null) {
+                jsonArray = new JSONArray();
             }
+            WritableArray array = TextileNode.convertJsonToArray(jsonArray);
             promise.resolve(array);
```
asutula: Previously, we were sending an array of `String`s here. Needed to make it consistent with the iOS version which sends an array of objects that will be properly converted to JS objects by RN. In this case, that object type is `WritableArray` (which is an array of `WritableMap`s). RN automatically converts them to JS objects.

android/app/src/main/java/com/textile/textilenode/TextileNode.java
```diff
@@ -289,7 +293,8 @@ public void pairDevice (String pkb64, Promise promise) {
     public void getFilePath(String uriString, Promise promise) {
         Uri uri = Uri.parse(uriString);
         try {
-            promise.resolve(RealPathUtil.getRealPath(reactContext, uri));
+            String path = RealPathUtil.getRealPath(reactContext, uri);
+            promise.resolve(path);
```
asutula: Noticed some weirdness here. Seems to work fine on a real device, but on the emulator, the path returned here eventually is identified as 'file doesn't exist'. Anyone seen that before?

asutula: @andrewxhill ^?

android/app/src/main/java/com/textile/textilenode/TextileNode.java
```diff
@@ -300,4 +305,109 @@ private static void emitDeviceEvent(String eventName, @Nullable WritableMap even
         // https://facebook.github.io/react-native/docs/native-modules-android.html#sending-events-to-javascript
         reactContext.getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class).emit(eventName, eventData);
     }
+
```
asutula: These are some utility methods to convert to/from `JSONObject` and `WritableMap/Array`. Probably a better place for them?

android/app/src/main/java/com/textile/textilenode/TextileNode.java
andrewxhill: haven't seen it, no... android only?

asutula: Yea android only.

asutula:
Closing this because it's included in #165

PR #163, opened 2018-06-28
wip: adds a login deeplink to grab referral code

andrewxhill:
will need to chat through this, still need to grab the token for joining the thread, but need to figure out where to store that off until signup is complete


andrewxhill:
pulled into my other work

PR #165, opened 2018-06-29
Textile image fixes/Android

asutula:
This provides a consistent `TextileImage` component API for Android and iOS. I also re-implemented the iOS version to make it more simple and similar to the Android implementation.


android/app/src/main/java/com/textile/textilenode/TextileImageTask.java
```diff
@@ -0,0 +1,52 @@
...
+import android.graphics.BitmapFactory;
+import android.os.AsyncTask;
+import android.util.Base64;
+import android.widget.ImageView;
+
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.uimanager.events.RCTEventEmitter;
+
+public class TextileImageTask extends AsyncTask<Void, Void, Bitmap> {
```
asutula: This guy allows us to load the image data from the Textile node on a background thread, and update the UI on the UI thread.

ios/NativeViews/TextileImageView.m
```diff
@@ -1,462 +1,91 @@
...
-  NSDictionary *dict = @{
-                         @"width": @(source.size.width),
-                         @"height": @(source.size.height),
-                         @"hash": source.hsh,
-                         @"path": source.path,
-                         };
-  return @{ @"source": dict };
-}
 
 @interface TextileImageView ()
```
asutula: Lots of deleted code in here because I had previously copied and modified Facebook's code, and now I just wrote it from scratch to make it our own and more simple.

android/app/src/main/java/com/textile/textilenode/TextileImageTask.java
andrewxhill: niceee

PR #175, opened 2018-07-12
Minimal thread management + sharing

asutula:
Here's a version that lets us see all threads, create new ones, and share photos to any of those threads.

Took the opportunity to convert some code to TypeScript. More work to come there.

This branch includes all the updates in #170, so we can just close that one and merge this one?


andrewxhill:
getting this

screen shot 2018-07-13 at 1 07 57 pm

andrewxhill:
been messing with this all day, gtg to me

Issue #177, opened 2018-07-15
Wait for pairing confirmation before confirming that the devices are paired


andrewxhill:
Currently in DevicesRedux, the state 'added' is reached as soon as the device is added to the wallet. But we need a step after this where the device is confirmed as being paired, this is later confirmed by a message from the external device saying it received the pair.

sanderpick:
I believe calling pair will either fail or succeed... as in, it's sync. We should expose the ability to _not_ fallback to offline message sending and do that here.

andrewxhill:
ahhh gotcha

PR #178, opened 2018-07-15
Adds deep link support and interfaces for new device pairing and new thread invites

andrewxhill:
This is all really close. We could merge it and forge on just knowing that we need to do something about this https://github.com/textileio/textile-mobile/issues/177 now or at some point before the desktop app goes out for testing.


andrewxhill:
Ah, this relies on the frameworks being generated by https://github.com/textileio/textile-go/pull/158

I didn't update the Android framework because that textile-go PR is still a WIP so we'll update both as soon as it is ready.... not ideal but keeping it moving

App/Containers/LoginScreen.js
```diff
@@ -56,6 +57,29 @@ class LoginScreen extends Component {
...
+  _handleOpenURLEvent (event) {
+    this._handleOpenURL(event.url)
+  }
+
+  _handleOpenURL (url) {
+    if (url) {
+      const data = DeepLink.getData(url)
+      const request = DeepLink.getParams(data.hash)
+      if (data.path === '/invites/new' && data.hash !== '' && request.referral) {
+        // TODO: need to store off the invite information until after the user signs up now
```
andrewxhill: Not sure the right way here ^

App/Containers/ThreadInvite.js
```diff
@@ -0,0 +1,154 @@
...
+  const nodeState = state.ipfs && state.ipfs.nodeState ? state.ipfs.nodeState.state === 'started' : false
+  return {
+    threads: state.threads && state.threads.threadItems ? state.threads.threadItems : [],
+    online: nodeState && online
+  }
+}
+
+const mapDispatchToProps = (dispatch) => {
+  return {
+    addThreadRequest: (name, id) => { dispatch(ThreadsAction.addThreadRequest(name, id)) },
```
andrewxhill: Just looking at textile-go work and realizing this is calling the wrong method. We want to call this method here, https://github.com/textileio/textile-go/pull/158/files#diff-95a702539b8bc16ea2eb14e093c5555aR290

andrewxhill: fixed now

PR #180, opened 2018-07-16
iOS Thread Invite Bridge

andrewxhill:


App/Containers/TextilePhotos.js
```diff
@@ -17,13 +19,14 @@ class TextilePhotos extends React.PureComponent {
...
     ) : params.thread
+
     return {
       headerTitle
     }
   }
 
   componentDidMount () {
-    this.props.navigation.setParams({ 
+    this.props.navigation.setParams({
```
asutula: Anything actually change in this file? Looks like some added imports that aren't used, etc?

andrewxhill: you might have been reviewing prior to my last commit, i had been messing around with it to gen an invite link for testing... should be reverted now

App/Redux/ThreadsRedux.ts
```diff
@@ -28,6 +28,24 @@ const actions = {
   }),
   refreshThreadsError: createAction('REFRESH_THREADS_ERROR', resolve => {
     return (error: Error) => resolve({ error })
+  }),
+  addExternalInvite: createAction('ADD_EXTERNAL_THREAD_INVITE', resolve => {
```
asutula: Just to be consistent, I'd name the actions that initiate the async sequence with 'request': `addExternalInviteRequest`, for example.

App/Redux/ThreadsRedux.ts
```diff
@@ -44,6 +62,15 @@ export type ThreadsState = {
     readonly threadId: string
     readonly error?: Error
   }
+  readonly links?: {
```
asutula: Nit here and for `invites` below... we're just tracking a single invite and link at a time, so get rid of the plural wording?

App/Redux/ThreadsRedux.ts
```diff
@@ -28,6 +28,24 @@ const actions = {
...
+  }),
+  acceptExternalInvite: createAction('ACCEPT_EXTERNAL_THREAD_INVITE', resolve => {
+    return (link: string) => resolve({ link })
+  }),
+  acceptExternalInviteSuccess: createAction('ACCEPT_EXTERNAL_THREAD_INVITE_SUCCESS', resolve => {
+    return () => resolve()
+  }),
+  acceptExternalInviteError: createAction('ACCEPT_EXTERNAL_THREAD_INVITE_ERROR', resolve => {
+    return (error: Error) => resolve({ error })
   })
```
asutula: Feels good to say this for once... There are tests for `ThreadsRedux`, so please update the tests to cover this new stuff ;)

App/Sagas/TextileSagas.ts
```diff
@@ -379,3 +379,24 @@ export function * refreshThreads () {
...
+    const link = yield call(TextileNode.addExternalThreadInvite, name, pubKey)
+    yield put(ThreadsActions.addExternalInviteSuccess(pubKey, link))
+  } catch (error) {
+    yield put(ThreadsActions.addExternalInviteError(error))
+  }
+}
+
+export function * acceptExternalInvite (action: ActionType<typeof ThreadsActions.acceptExternalInvite>) {
+  const { link } = action.payload
+  console.log('ACT', action, action.payload)
```
asutula: Let's clean up all the console logs if this is working.

andrewxhill: make sure you refresh

TextileNode.ts
```diff
@@ -66,6 +66,14 @@ export default {
     return threads
   },
 
+  addExternalThreadInvite: async function (name: string, pubKey: string): Promise<string> {
```
asutula: Just learning here for when I integrate this into the app... What is the string being returned?

asutula:
Awesome, about ready for this on my end, so excited to get it merged!

TextileNode.ts
andrewxhill: it will be most of the link you'll need. so it should look like,

`https://www.textile.photos/invites/new#invite=QmTdt36G6RXktJ9UypUftiqGdogidNBmxwwNcHW39Jg3WB&key=17TxeCbotq7wzm9AsRpcRIH817617TxeCCn1iOQjQCyz&name=TestTest`

i think @sanderpick was going to change the `invite` key, but it shouldn't matter since the RN will never need to get into that. I'd also like RN to add a final parameter to the url, `&from=andrewxhill` with some identifier of the invitee so that we can use it to give context in the message of the invited

App/Redux/ThreadsRedux.ts
andrewxhill: done

App/Redux/ThreadsRedux.ts
andrewxhill: will do

PR #183, opened 2018-07-17
Invite Dialog

asutula:


App/Containers/ThreadInvite.js
```diff
@@ -26,6 +26,7 @@ class ThreadInvite extends React.PureComponent {
   }
 
   cancel = () => {
+    // TODO: remove thread takes thread name... what is key?
     this.props.removeThreadRequest(this.state.key)
```
asutula: @andrewxhill this function to remove the thread needs to take the thread name (until we fix that API)... wasn't sure what `this.state.key` represents, but let's make sure we're passing thread name here?

andrewxhill: i think we'll have a fix for it later today from the go side, so maybe just let it be until framework update

asutula:
Ok, besides the question i left for @andrewxhill, this is good to go.

App/Containers/ThreadInvite.js
asutula: cool

andrewxhill:
Let's do it! Team Thread in 3. 2. 1.... CircleCI com'on! 3.. 2..

App/Sagas/TextileSagas.ts
```diff
@@ -389,6 +389,11 @@ export function * addExternalInvite (action: ActionType<typeof ThreadsActions.ad
   }
 }
 
+export function * presentShareInterface(action: ActionType<typeof ThreadsActions.addExternalInviteSuccess>) {
+  const { link } = action.payload
+  yield call(Share.share, { title: 'Join my thread on Textile!', message: link })
+}
```
andrewxhill: link should be appended with "&from=some-screen-name" before this point.

PR #194, opened 2018-07-18
Invite frameworks

andrewxhill:


App/Redux/ThreadsRedux.ts
```diff
@@ -30,16 +30,22 @@ const actions = {
...
   }),
   addExternalInviteSuccess: createAction('ADD_EXTERNAL_THREAD_INVITE_SUCCESS', resolve => {
-    return (pubKey: string, link: string) => resolve({ pubKey, link })
+    // TODO this is the return from Go, we need a ThreadExternalInvite type
+    // ExternalInvite{
+    //   Id:     string,
+    //     Key:     string,
+    //     Inviter: string,
+    // }
+    return (invite: TextileTypes.ThreadExternalInvite) => resolve({ invite })
```
andrewxhill: @asutula this could use a new type for the response here

Issue #206, opened 2018-07-20
Accepting an invite before signing up will not complete the thread invite after signup


andrewxhill:
We'll want this to work smoothly for DWeb

andrewxhill:
will make this a priority today

PR #209, opened 2018-07-20
Stores deeplink in threads to be processed post-login

andrewxhill:
fixes https://github.com/textileio/textile-mobile/issues/206


andrewxhill:
Okay, take a look. It only needs to run after the onboarding is complete so I added it as a task to run after success there, instead of running on node start (which happens too early in the onboarding case anyway)

PR #211, opened 2018-07-20
Password error

andrewxhill:


App/Redux/AuthRedux.ts
```diff
@@ -81,7 +81,11 @@ export function reducer (state: AuthState = initialState, action: AuthAction): A
     case getType(actions.updateFormType):
       return { ...state, formType: action.payload.formType }
     case getType(actions.updateFormValue):
-      return { ...state, formValue: action.payload.formValue }
+      const values = action.payload.formValue
+      const error = values.password && values.password.length > 28 ? 'React Native bug limits passwords to less than 28 characters' : undefined
```
asutula: wow

asutula:
Did a quick test. Seems like this error happens whenever the password string is longer than the text input is wide (I was testing with the iPhone 6 which can fit 30-somthing chars). Attaching a screen cap of the stack while it crashes. Might have more luck googling for words from that stack.
screen shot 2018-07-20 at 18 49 16

asutula:
Just tested a long password with the SB screens... no problem. So this problem is probably with the form library we were using. Can close this I think.

PR #221, opened 2018-07-23
SB onboarding

asutula:
Bare minimum to get the onboarding and sign up/in screens in. There are things that need to be changed/added. Figured we can test it out and ticket up any needed changes/enhancements for SB or ourselves.


asutula:
Don't think we should merge this one until we're totally happy with it. Until then, what we already have in the app is better.

PR #233, opened 2018-07-24
updates to bridges for textileio/textile-go#195

andrewxhill:


TextileNode.ts
```diff
@@ -5,6 +5,7 @@ import {
   DeviceEventEmitter,
 } from 'react-native'
 import * as TextileTypes from './App/Models/TextileTypes'
+import { CafeTokens } from './App/Redux/AuthRedux'
```
asutula: Better to put `CafeTokens` in `TextileTypes.ts`... Trying to consolidate all the types returned by the API in one place so we can eventually provide those definitions externally (from a release of the TextileNode native module package).

asutula:
I can't figure out what's wrong with the Android CI... It's a mystery for now, so ignore Android CI.

TextileNode.ts
andrewxhill: done

andrewxhill:
i think he might have just been referring to CafeTokens json versus old Token strong.

PR #235, opened 2018-07-26
Photo views

andrewxhill:


App/Navigation/PhotosNavigation.js
```diff
@@ -34,6 +40,20 @@ const PhotosNav = StackNavigator(
   }
 )
 
+const ThreadPhotoDetailViewer = StackNavigator(
+  {
+    ViewPhoto: {
```
asutula: Like this, `ThreadPhotoDetail` is in it's own `StackNavigator` which is presented modally when the user selects an image. The SB designs make it more of a 'push' onto and existing nav stack.

asutula: Ah, i can't comment on a line that hasn't changed, but look at line 24 above. You can add the `ThreadPhotoDetail` screen there, and it will be pushed onto the `PhotosNav` `StackNavigator`, more consistent with the SB designs.

PR #236, opened 2018-07-27
Image Processing

asutula:
Includes:

- New Typescript redux for camera roll and uploading images state
- Re-written `photosTask`, which is similar to the old, but more simple and easy to understand
- Better upload retry handling, waiting until the next 'tick' to retry instead of retrying immediately after a failure
- Better wiring of `photosTask` into redux events. It should now ignore any additional `startNodeSuccess` triggers if the `photosTask` is already in flight.

TODO in a later PR:
I had some work around improving the triggering of creating/starting the node, and therefore, indirectly this could effect triggering the `photosTask`... Needs to simmer a little longer.


asutula:
Oh, image upload progress pies are part of the 'verboseUi' preference. Oh, and I added a camera roll querying status label in the same verbose ui as where the node status is displayed. It's right next to it.

asutula:
Oh this also includes the updated background upload native module with the fix so it uses file based uploads (which are actually background uploads). Should help lots with background upload errors we've been seeing.

PR #263, opened 2018-07-29
point to working version of svg lib

andrewxhill:


asutula:
Wanna re-enable the android build in .circleci/config.yml and see what happens?

PR #270, opened 2018-07-30
New Framework & Profile images

andrewxhill:


App/Config/ReduxPersist.js
```diff
@@ -66,14 +66,14 @@ const REDUX_PERSIST = {
   reducerVersion: '1.0',
   storeConfig: {
     key: 'primary',
-    version: 4,
+    version: 5,
```
asutula: I don't think you need to increment this if you're not providing any migration.

andrewxhill:
@asutula I added 'contacts' to the persist whitelist, take a look and see if okay with that

App/Redux/ThreadsRedux.ts
```diff
@@ -213,6 +213,7 @@ export function reducer (state: ThreadsState = initialState, action: ThreadsActi
 }
 
 export const ThreadsSelectors = {
+  threads: (state: any) => state.threads,
```
asutula: Dup?

App/Config/ReduxPersist.js
andrewxhill: undone

App/Redux/ThreadsRedux.ts
andrewxhill: gone

App/Sagas/TextileSagas.ts
```diff
@@ -97,9 +96,14 @@ export function * handleProfilePhotoSelected(action: ActionType<typeof UIActions
     }
     const blockId: string = yield call(TextileNode.addPhotoToThread, addResult.id, addResult.key, defaultThread.id)
     yield put(UploadingImagesActions.addImage(addResult.archive.path, addResult.id, 3))
+
+
     yield put(TextileNodeActions.getPhotoHashesRequest(defaultThread.id))
 
-    //TODO: Call some TextileNode method to add profile picture
+    // set it as our profile picture
+    yield call(TextileNode.setAvatarId, addResult.id)
```
asutula: Nice.

App/Config/ReduxPersist.js
```diff
@@ -73,7 +73,7 @@ const REDUX_PERSIST = {
     // blacklist: ['nav', 'ipfs', 'auth', 'ui'],
     // Optionally, just specify the keys you DO want stored to persistence.
     // An empty array means 'don't store any reducers' -> infinitered/ignite#409
-    whitelist: ['preferences', 'uploadingImages', 'cameraRoll'],
+    whitelist: ['preferences', 'uploadingImages', 'cameraRoll', 'contacts'],
```
asutula: Yea looks good for now. Doesn't seem like a good long term plan to rely on this persistence if this is the only place we're keeping track of contacts (like this should somehow be stored in a decentralized way?)

PR #273, opened 2018-07-30
lots of changes in little adds

andrewxhill:


App/Containers/TextilePhotos.js
```diff
@@ -187,7 +188,11 @@ const mapStateToProps = (state, ownProps) => {
     thread = state.threads.threads.find(thread => thread.id === threadId)
   }
 
-  const threadName = thread ? thread.name : undefined
+  // I saw a really weird state where thread was all undefined....
+  // seems like we should show a loading state if that ever happens.
+  // at the very least i put the user on the default screen instead of a
+  // blank Thread screen
+  const threadName = thread ? thread.name : 'default'
```
andrewxhill: @asutula potential change here of significance. i can't see when a thread.name wouldn't exist but in a broken state i at least try to put them in the wallet view instead of a thread view (with broken back button etc)

asutula: Well there's a period of time when `thread` is `undefined`. This happens quickly while the thread data is being async fetched from the node. Yea, we should display a loading indicator or something. 

I'm not sure forcing a thread name of `default` really fixes anything, it may even be inconsistent with the `threadId` at the same time.

App/SB/views/PhotoDetail/index.js
```diff
@@ -45,7 +42,9 @@ class PhotoDetail extends Component {
             <Image style={styles.toolbarAddIcon} source={require('./statics/icon-add.png')}/>
           </TouchableOpacity>
           <Image style={styles.toolbarDownloadIcon} source={require('./statics/icon-download.png')}/>
-          <Image style={styles.toolbarShareIcon} source={require('./statics/icon-share.png')}/>
+          <TouchableOpacity onPress={params.getPublicLink}>
```
asutula: There's a nice library we were using called react-native-header-buttons that makes this a lot nicer, you wouldn't have to be "creating" "buttons" by using `Image` and `TouchableOpacity`. Not sure if you want to change now, could do it later.

asutula:
ah ok already merged

App/SB/views/PhotoDetail/index.js
andrewxhill: ah nice. will start using

PR #309, opened 2018-08-06
TS Redux Integration

asutula:
Brought the TS reducer stuff full circle by creating a type safe redux store as well as a typed `RootState` and `RootAction` that can now be used in React components. Started converting a couple of the low level components (`App`, `RootContainer`, and `OnboardingCheck`) over to TS.


App/Config/ReduxPersist.ts
```diff
@@ -0,0 +1,82 @@
+import { AsyncStorage } from 'react-native'
+import { createMigrate, PersistConfig, MigrationManifest } from 'redux-persist'
+
+const migrations: MigrationManifest = {
```
asutula: Still need to figure out the right way to be able to use TS types during data migration. For now, just casting to `any` let's us pretend it's plain JS objects.

App/Containers/OnboardingCheck.tsx
```diff
@@ -0,0 +1,34 @@
...
+  render () {
+    return (
+      <View style={{flex: 1}}>
+        <ActivityIndicator />
+      </View>
+    )
+  }
+}
+
+const mapStateToProps = (state: RootState) => {
```
asutula: Example of using `RootState` here.

App/Containers/OnboardingCheck.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import { ActivityIndicator, View } from 'react-native'
+import { connect } from 'react-redux'
+import { NavigationScreenProps } from 'react-navigation'
+
+import { RootState } from '../Redux/Types'
+
+interface Props extends NavigationScreenProps<{}> {
```
asutula: Example of defining components props type. In this case, we also adopt NavigationScreenProps because this screen can live inside of a navigator.

App/Containers/PairingView.js
```diff
@@ -146,8 +146,8 @@ class PairingView extends React.PureComponent {
 }
 
 const mapStateToProps = state => {
-  const online = state.ipfs && state.ipfs.online && state.ipfs.online ? state.ipfs.online : false
-  const nodeState = state.ipfs && state.ipfs.nodeState ? state.ipfs.nodeState.state === 'started' : false
+  const online = state.textileNode && state.textileNode.online && state.textileNode.online ? state.textileNode.online : false
```
asutula: Lots of file changes here just renaming the state key `ipfs` to `textileNode`.

typings/react-native-background-task/index.d.ts
```diff
@@ -0,0 +1,21 @@
+declare module 'react-native-background-task' {
```
asutula: In this `typings` folder, we can add TS definitions for modules that don't have TS definitions available.

PR #314, opened 2018-08-07
Photo picker

andrewxhill:


App/Containers/AddCaptionScreen.js
```diff
@@ -21,7 +22,7 @@ class AddCaptionScreen extends React.Component {
   static navigationOptions = ({ navigation }) => {
     const params = navigation.state.params || {}
     return {
-      headerTitle: undefined,
+      headerTitle: 'Photo Caption',
```
andrewxhill: Based on some feedback watching people onboard, they were confused by the large title hanging under the back arrow. they wanted to be able to click it, not realizing the form was below it. so for now, i've just moved the title up into the top bar.

App/Containers/AddCaptionScreen.js
```diff
@@ -58,29 +59,45 @@ class AddCaptionScreen extends React.Component {
...
-    this.props.close()
-    this.props.navigation.goBack()
+  _close () {
+    const params = this.props.navigation.state.params
+    if (params.photo) {
+      // from the wallet
+      this.props.close()
+    } else {
+      // from the photo picker
+      this.props.cancelShare(params.threadId, params.image)
```
andrewxhill: just allows this page to arise from sharing from Wallet or from sharing from cameraroll

App/Containers/TextilePhotos.js
```diff
@@ -15,14 +15,11 @@ import BottomDrawerList from '../SB/components/BottomDrawerList'
 class TextilePhotos extends React.PureComponent {
   constructor (props) {
     super(props)
-    this.state = {
-      showDrawer: false
-    }
   }
```
andrewxhill: just did a bunch of cleanup in this container. it is only ever called in the default thread now so lots of code removed

App/Redux/CameraRollRedux.ts
```diff
@@ -30,13 +51,16 @@ export type QueriedPhotosMap = {
 export type CameraRollState = {
   readonly initialized: boolean
   readonly querying: boolean
-  readonly queriedPhotos: QueriedPhotosMap
+  readonly queriedPhotos: QueriedPhotosMap,
+  // A map of threads and their pending images from pendingImageShares
+  readonly pendingShares: { [index:string] : Array<TextileTypes.SharedImage> }
```
andrewxhill: pendingShares only hang around while the image is being processed prior to upload. once it is added to the upload queue, this guy gets removed

App/Redux/CameraRollRedux.ts
```diff
@@ -85,6 +109,95 @@ export function reducer (state: CameraRollState = initialState, action: CameraRo
...
+      ps.push(partial)
+
+      return {
+        ...state,
+        pendingShares: {
+          ...state.pendingShares,
+          [action.payload.threadId]: ps
+        }}
+    case getType(actions.remotePinSuccess):
+    case getType(actions.imagePinError):
```
andrewxhill: Not a very exciting Error handler, just removes it like a success. I added a small error 'Alert' in the thread view so that any error will tell the user something and move along

App/SB/views/ThreadsDetail/index.js
```diff
@@ -106,17 +113,43 @@ class ThreadsEdit extends React.PureComponent {
...
+      return {height: 1, backgroundColor: 'transparent', flex: 1.0 - this.props.progress}
+    }
+  }
+
   render () {
     return (
       <View style={styles.container}>
-        <ScrollView style={styles.contentContainer}>
-          {this.props.items.map((item, i) => <ThreadDetailCard key={i} last={i === this.props.items.length - 1} {...item} onSelect={this.onPhotoSelect()}/>)}
+        {this.props.showProgress && <View style={{height: 1, flexDirection: 'row', padding: 0, margin: 0}}>
```
andrewxhill: adds a small progress feedback when an image is shared. will post ss in slack

App/Sagas/TextileSagas.ts
```diff
@@ -47,7 +48,7 @@ export function * signUp (action: ActionType<typeof AuthActions.signUpRequest>)
     yield put(AuthActions.getTokensSuccess(tokens))
     // TODO: Put username into textile-go for addition to metadata model
     yield put(AuthActions.signUpSuccess())
-    yield call(NavigationService.navigate, 'Permissions')
+    yield call(NavigationService.navigate, 'ProfilePic')
```
andrewxhill: no longer requesting permissions up top

App/Containers/AddCaptionScreen.js
asutula: Makes sense

App/Redux/CameraRollRedux.ts
```diff
@@ -85,6 +109,95 @@ export function reducer (state: CameraRollState = initialState, action: CameraRo
...
+        return state
+      }
+      const existing = state.pendingShares[action.payload.threadId].find(img=>img.origURL === action.payload.image.origURL)
+      if (!existing || existing.caption !== undefined) {
+        return state
+      }
+      existing.caption = action.payload.caption
+
+      const thread = state.pendingShares[action.payload.threadId].filter((img) => img.origURL !== action.payload.image.origURL)
+      thread.push(existing)
```
asutula: This is a bit confusing, but I think I get it. You want to replace the old image with the new version (updated caption). Would be easier with a `map` operation I think.

App/Redux/CameraRollRedux.ts
```diff
@@ -85,6 +109,95 @@ export function reducer (state: CameraRollState = initialState, action: CameraRo
...
+        return state
+      }
+      const partial = state.pendingShares[action.payload.threadId].find(img=>img.origURL === action.payload.image.origURL)
+      if (!partial || partial.addResult !== undefined) {
+        return state
+      }
+      partial.addResult = action.payload.addResult
+
+      const ps = state.pendingShares[action.payload.threadId].filter((img) => img.origURL !== action.payload.image.origURL)
+      ps.push(partial)
```
asutula: Same here, cleaner with a `map`

App/Redux/CameraRollRedux.ts
```diff
@@ -85,6 +109,95 @@ export function reducer (state: CameraRollState = initialState, action: CameraRo
...
+        return state
+      }
+
+      const filteredShares = state.pendingShares
+      const newArray = filteredShares[action.payload.threadId].filter((img) => img.origURL !== action.payload.image.origURL)
+      if (newArray.length > 0) {
+        filteredShares[action.payload.threadId] = newArray
+      } else {
+        // remove the key from the object if array=0
+        delete filteredShares[action.payload.threadId]
```
asutula: Cool, what does `delete` do!?

asutula: I guess it's obvious, but cool, didn't know about that.

App/Redux/CameraRollRedux.ts
```diff
@@ -85,6 +109,95 @@ export function reducer (state: CameraRollState = initialState, action: CameraRo
...
+        return {...state, pendingShares}
+      } else {
+        return {
+          ...state,
+          pendingShares: {
+            ...state.pendingShares,
+            [action.payload.threadId]: filteredArray
+          }
+        }
+      }
```
asutula: Lots of new logic in here, hard to follow. Would be good to update the tests ;)

App/SB/views/ThreadsDetail/index.js
```diff
@@ -174,29 +206,51 @@ const mapStateToProps = (state, ownProps) => {
...
+    if (firstShare.caption) {
+      progress = 0.3
+      if (firstShare.addResult) {
+        progress = 0.6
+        if (state.uploadingImages.images[firstShare.addResult.id]) {
+          progress = 0.8
+        }
+      }
+    }
+  }
```
asutula: Cool. The first two steps are almost instantaneous, right? I wonder if just displaying the actual image upload progress would be more meaningful (kinda like our pie views before).

App/Sagas/TextileSagas.ts
```diff
@@ -564,6 +565,120 @@ export function * refreshThreads () {
   }
 }
 
+export function * showImagePicker(action: ActionType<typeof UIActions.showImagePicker>) {
+  const { threadId } = action.payload
+
+  // Turn ImagePicker into a Promise
+  const pickerPromise = new Promise((resolve, reject) => {
```
asutula: This code already exists in `CameraRoll.ts` if you want to use it. Just a simple function you can call that returns a `Promise<{ uri: string, data: string}>`

App/Redux/CameraRollRedux.ts
andrewxhill: adding a few now

App/SB/views/ThreadsDetail/index.js
andrewxhill: this is pre-upload... the image even shows in the thread before remote pin, so no concrete requirement to show pie. so if you are looking at the ui there is 1 second where nothing appears, it feels a lot better to see progress, otherwise it just feels a bit laggy for no stated reason. give it a try

App/Sagas/TextileSagas.ts
andrewxhill: ah, perf

App/Sagas/TextileSagas.ts
```diff
@@ -564,6 +565,120 @@ export function * refreshThreads () {
...
+    yield put(UploadingImagesActions.addImage(addResult.archive.path, addResult.id, 3))
+
+    // Begin the image upload to the Cafe pinner
+    yield uploadFile(
+      addResult.id,
+      addResult.archive.path
+    )
+
+    // Remove the image from the CameraRoll actions
+    yield put(CameraRollActions.remotePinSuccess(threadId, image))
```
asutula: At this point, the image hasn't finished uploading. If `CameraRollActions.remotePinSuccess` is intended to fire once the image is done uploading, it needs to happen once we're notified the upload is complete, similar to how we are doing it in `TextileManager`, firing `UploadingImagesActions.imageUploadComplete`.

andrewxhill: Right, the intention here was more, remotePinInitSuccess. As in, the problem is no longer the camerarolls, now it's over to the uploader

App/SB/views/ThreadsDetail/index.js
asutula: nice, makes sense.

App/Services/CameraRoll.ts
```diff
@@ -71,3 +82,34 @@ export async function chooseProfilePhoto(): Promise<{ uri: string, data: string}
     })
   })
 }
+
+export async function choosePhoto(): Promise<PickerImage> {
```
asutula: Ok this is fine for now, but we should consolidate these two functions into one, they're basically the same.

PR #316, opened 2018-08-07
Remove react-native-permissions

asutula:
Also disable location monitoring for now.


asutula:
Going to merge so we can get a Testflight out. Objections?

PR #317, opened 2018-08-07
cameraroll migration

andrewxhill:
this param would be missing, could cause issues if not handled nicely by redux?


andrewxhill:
going to get this in

PR #319, opened 2018-08-08
Fix release mode JS exception

asutula:
Was causing all sorts of problems, including how we saw sign in not working.

### DIFF
--- a/App/Redux/PreferencesRedux.ts
+++ b/App/Redux/PreferencesRedux.ts
@@ -1,5 +1,5 @@
 import { createAction, ActionType, getType } from 'typesafe-actions'
-import TextileTypes from '../Models/TextileTypes'
+import { Profile } from '../Models/TextileTypes'
 
 const actions = {
   onboardedSuccess: createAction('ONBOARDED_SUCCESS', resolve => {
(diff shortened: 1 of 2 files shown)


App/Redux/PreferencesRedux.ts
```diff
@@ -1,5 +1,5 @@
 import { createAction, ActionType, getType } from 'typesafe-actions'
-import TextileTypes from '../Models/TextileTypes'
+import { Profile } from '../Models/TextileTypes'
```
asutula: Not the problem, but a compile time error I saw.

App/Sagas/TextileSagas.ts
```diff
@@ -180,14 +179,11 @@ export function * handleNewAppState (action: ActionType<typeof TextileNodeAction
   const { previousState, newState } = action.payload
   console.log('handleNewAppState', previousState, newState)
   if (previousState.match(/default|unknown/) && newState === 'background') {
-    Reactotron.logImportant('launched into background')
```
asutula: These were the problem. `Reactotron` isn't available in a release build, doh.

PR #320, opened 2018-08-08
Settings & notifications

andrewxhill:


andrewxhill:
Will need to do something with the Notifications page. I can stub in some basic ones that are more or less static (mnemonic created, node created, node started) and then we can build around it as a later pr

App/Navigation/AppNavigation.js
```diff
@@ -4,6 +4,7 @@ import TextileManager from '../Containers/TextileManager'
 import OnboardingCheck from '../Containers/OnboardingCheck'
 import AndrewPairingInvite from '../SB/views/AndrewDeviceInvite'
 import AndrewThreadInvite from '../SB/views/AndrewThreadInvite'
+import AccountNavigation from './AccountNavigation'
```
asutula: Don't need this here.

App/Navigation/PhotosNavigation.js
```diff
@@ -2,21 +2,23 @@ import React from 'react'
 import { StackNavigator, TabNavigator, TabBarBottom } from 'react-navigation'
 import { Image } from 'react-native'
 import TextilePhotos from '../Containers/TextilePhotos'
-import InfoView from '../Containers/InfoView'
 
 import styles, {headerTintColor} from './Styles/NavigationStyles'
+import AccountNavigation from './AccountNavigation'
```
asutula: Unused I think.

App/Redux/AuthRedux.ts
```diff
@@ -20,6 +20,9 @@ const actions = {
   logInRequest: createAction('LOG_IN_REQUEST', resolve => {
     return (username: string, password: string) => resolve({ username, password })
   }),
+  logOutRequest: createAction('LOG_OUT_REQUEST', resolve => {
```
asutula: Don't see any 'log out succes' action, but do we need it so we can clear out the `CafeTokens` stored in `AuthRexux`?

App/Navigation/PhotosNavigation.js
andrewxhill: It is, it's part of the root nav at the bottom

App/Redux/AuthRedux.ts
andrewxhill: I didn't complete it, figured it needs its own PR later. I commented out the Lock Screen button in the settings for now. The reason was that we need a few things to make it work,

- [ ] Goodbye screen (optional)
- [ ] When logged out after loggin in, we might want to skip those three intro screens before the login
- [ ] We currently force the user to retake their profile pic after any login, we prob want to skip that or at least ensure it works

andrewxhill:
hmm.. just merged master into it and it doesn't like something. looking into it

andrewxhill:
okay. let's do it

PR #321, opened 2018-08-08
Cleans up a lot of the un-clickable buttons

andrewxhill:


andrewxhill:
also gtg

asutula:
What do you think about using `react-native-header-buttons`? Not sure if you've tried it out, but it worked nicely in previous versions of the app.

andrewxhill:
ah shoot, forgot you had mentioned those before. let's get this guy in and i'll go back and replace next time around

asutula:
Are there any changes in this branch we actually need to keep then? Or can we just close this?

andrewxhill:
dunno, i'd say let's put it in because it makes the app more usable already... (there are non-header buttons fixed here too) and then i can replace with that follow-up work

asutula:
I don't know about merging this because it will definitely lead to a lot of unused code, mostly styling code, sitting around once we switch to to header buttons.

asutula:
ok, let's just be sure to actually clean up later then

PR #322, opened 2018-08-08
Notifications redux

andrewxhill:


App/Models/TextileTypes.ts
```diff
@@ -22,6 +22,13 @@ export interface Devices {
   items: Device[]
 }
 
+export interface Notification {
```
asutula: I originally made the `TextileTypes` file to keep track of the model types that are used with `textile-go`... Let's make a point of cleaning that up, makes it easier to deal with the framework update process.

App/Redux/RootReducer.ts
```diff
@@ -17,6 +17,7 @@ export default combineReducers({
   contacts: contactsReducer,
   devices: devicesReducer,
   preferences: prefrencesReducer,
+  notifications: notificationsReducer,
```
asutula: Cool, think you need to add the 'notifications' key to the persistence whitelist as well.

App/Models/TextileTypes.ts
andrewxhill: cool. pretty sure this guy will live here regardless, it's the UI-based reshaping of notifications. making it so i can easily style the feed list item by type of notification, stuff that go isn't going to care about

App/Redux/RootReducer.ts
andrewxhill: done

App/Models/TextileTypes.ts
```diff
@@ -22,6 +22,13 @@ export interface Devices {
...
+  category: 'node' | 'devices' | 'threads' | 'content'
+  type: string
+  date: number
+  payload?: any
+}
```
asutula: Seems like we'd want to track a 'read' or 'unread' status.

App/Redux/NotificationsRedux.ts
```diff
@@ -0,0 +1,33 @@
+import { createAction, ActionType, getType } from 'typesafe-actions'
+import * as TextileTypes from '../Models/TextileTypes'
+
+const actions = {
+  newNotification: createAction('NEW_NOTIFICATION_REQUEST', resolve => {
```
asutula: Good start I think. I could imagine some other actions we'd want:

1. Acknowledge notification - Fired it the user taps or opens a notification, toggles the status to 'read'
2. Acknowledge all notifications - Fired in a situation like, user opens the notifications screen and is there for at least 5 seconds, and then navigates away from the notifications screen. Marks all unread notifications as read.

PR #341, opened 2018-08-10
Wallet view

andrewxhill:
Adds a list/grid view toggle. bit more padding. closer to desired layout


App/Redux/PreferencesRedux.ts
```diff
@@ -34,11 +37,13 @@ export type PreferencesState = {
   publicKey?: string
   profile?: Profile
   pending?: string
+  uiSettings: {[index: string]: string}
```
asutula: We loose the benefit of TS by making this object be any `string` -> `string`. We always have to know which settings we support, so might as well be explicit about it and type it, similar to the `verboseUi` setting. Related, if we want to group 'ui' settings, i guess we should migrate `verboseUi` into it.

App/Redux/__tests__/PreferencesRedux.ts
```diff
@@ -32,4 +32,10 @@ describe('preferences stories', () => {
       expect(state.mnemonic).toEqual(mnemonic)
     })
   })
-})
\ No newline at end of file
+  describe('grid view', () => {
+    it('should update mnemonic', () => {
```
asutula: Copy pasta typo

App/Components/PhotoGrid.js
```diff
@@ -4,7 +4,8 @@ import {
   Image,
   Text,
   FlatList,
-  TouchableOpacity
```
asutula: Any reason we want to update `PhotoGrid.js` instead of put the component from SB in the SB folder like we were doing for many of the others? Just trying to make it easier to clean up later when we want to remove all unused/old code.

asutula:
Seeing this when i add some photos to the wallet:
screen shot 2018-08-10 at 16 11 58
Is that ok for now or do we want to work on it?

asutula:
Also notices usernames aren't displaying in the thread timeline screen. Intentional because of FW updates?

andrewxhill:
usernames are pending the latest fw update

App/Components/PhotoGrid.js
andrewxhill: Just didn't see a ton of gain going that direction on this one

PR #342, opened 2018-08-11
Frameworks 228 update

andrewxhill:
- Metadata is now only requested when needed -- when photo sizes are required mostly
- refreshMessages is now called whenever user pull refreshes the ThreadsList or ThreadsDetail scrollviews and before either of those threads are rendered

refreshMessages should be called sometimes during the background routines, but not certain where that fits in.

### DIFF
--- a/App/Redux/UIRedux.ts
+++ b/App/Redux/UIRedux.ts
@@ -11,6 +11,9 @@ const actions = {
   selectProfilePicture: createAction('SELECT_PROFILE_PICTURE', resolve => {
     return (uri: string) => resolve({ uri })
   }),
+  viewThreadRequest: createAction('VIEW_THREAD_REQUEST', resolve => {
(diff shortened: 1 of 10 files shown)


andrewxhill:
doooo it

andrewxhill:
while aaron's away, merge all day

App/Redux/UIRedux.ts
```diff
@@ -52,6 +55,15 @@ const actions = {
   }),
   dismissImagePickerError: createAction('DISMISS_IMAGE_PICKER_ERROR', resolve => {
     return () => resolve()
+  }),
+  refreshMessagesRequest: createAction('REFRESH_MESSAGES_REQUEST', resolve => {
```
asutula: Was trying to keep any redux actions that query the TextileNode in `TextileNodeRedux`.

PR #344, opened 2018-08-12
Navigation Update + Icons

asutula:
- Updated to `react-navigation` 2.x + TypeScript typings
- Refactored app navigation into a nice directory structure to make it easier to understand the nested navigator structure
- All navigators migrated to TypeScript
- Created an icon font for all toolbar and header icons
- API for using the icon font with `react-native-vector-icons`, like how we used to use `EvilIcon`, `IonIcons`, etc
- Updated all nav bars and the tab bar to use `react-native-header-buttons` and `react-native-vector-icons` with our icon font
- Had to remove the tall header from a few screens that had them for now. That was the easiest way to have consistency throughout all the screens immediately, while we develop the real solution for tall headers.


asutula:
Be sure to `yarn install` if you run this.

asutula:
Oh, I also made all icon buttons in the header black for now, and any text button in the header blue. There are many different colored icon buttons in the headers of the SB designs, and it wasn't making much sense to me, so this seemed good and simple for now. Can revisit for sure.

andrewxhill:
this is great

PR #364, opened 2018-08-14
Adds image height util to enable 0 height images

andrewxhill:
You'll need this for framework textile-go#234 but i didn't bother including the framework this time around since you'll update it on 236 anyway


andrewxhill:
5 min on this one

asutula:
What's the reason behind this? To support old images that don't have aspect ratio in the block? Something else? Just want to understand.

asutula:
@andrewxhill want to fix this and i'll get it merged?

Issue #346, opened 2018-08-13
Delete / remove photo


sanderpick:

sanderpick:
(needs a go API) - https://github.com/textileio/textile-go/issues/171

PR #370, opened 2018-08-15
Delete photos

andrewxhill:
fixes https://github.com/textileio/textile-mobile/issues/346

### DIFF
--- a/App/Config/ReduxPersist.ts
+++ b/App/Config/ReduxPersist.ts
@@ -90,13 +90,26 @@ const migrations: MigrationManifest = {
         }
       }
     }
+  },
+  7: (persistedState) => {
+    const state = persistedState as any
+    return {
+      ...state,
+      preferences: {
+        ...state.preferences,
+        tourScreens: {
+          ...state.preferences.tourScreens,
+          threads: true
+        }
+      }
+    }
   }
 }
 
 const persistConfig: PersistConfig = {
   key: 'primary',
   storage: AsyncStorage,
-  version: 6,
+  version: 7,
   whitelist: ['preferences', 'uploadingImages', 'cameraRoll', 'notifications'],
   migrate: createMigrate(migrations, { debug: false })
 }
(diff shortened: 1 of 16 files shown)


App/Config/ReduxPersist.ts
```diff
@@ -90,13 +90,26 @@ const migrations: MigrationManifest = {
...
+  },
+  7: (persistedState) => {
+    const state = persistedState as any
+    return {
+      ...state,
+      preferences: {
+        ...state.preferences,
+        tourScreens: {
+          ...state.preferences.tourScreens,
+          threads: true
```
andrewxhill: Converted the threads onboarding screen into the same as the wallet one. now that threads are the default view, it was creating a flash of onboarding every time it loaded with the old way anyway

App/SB/views/PhotoDetail/index.js
```diff
@@ -194,19 +218,21 @@ const mapStateToProps = (state, ownProps) => {
...
   const path = thread.name === 'default' ? '/photo' : '/thumb'
+  const source = photo ? {url: 'file://' + photo.id + '.png'} : {url: 'file://.png'}
+
   return {
     photo,
-    date: photo.date.split('T')[0],
-    key: photo.id + path,
-    source: {url: 'file://' + photo.id + '.png'}, // <-- in case RN uses to know things
+    date: photo && photo.date ? photo.date.split('T')[0] : '',
+    key: photo && photo.id ? photo.id + path : path,
```
andrewxhill: all this is just making it safe the if the view doesn't unmount in time before the image remove request is done that it wont error out

andrewxhill:
okay, updated android bridge. gtg

App/Redux/PreferencesRedux.ts
```diff
@@ -45,7 +45,8 @@ export const initialState: PreferencesState = {
   onboarded: false,
   verboseUi: false,
   tourScreens: {
-    wallet: true
+    wallet: true,
+    threads: true
```
asutula: I don't understand why these would initially be `true`. Seems that the tour UI would never show for new users then. Maybe I'm missing something.

andrewxhill: true == yes, show it

PR #375, opened 2018-08-15
Image sharing update

asutula:
WIP: Still need to get the UI for visualizing the processing images in the thread timeline.

But, it's to the point where we should start discussing the code.

- Cleans up image sharing data flow
- Adds new `ProcessingImagesRedux` that tracks state of images before they are done sharing to a thread.
- Reverses the order of processing, now we add to IPFS, upload, add to wallet, share to thread


asutula:
Ah need to update the tests as well, they're failing now

App/Containers/AddCaptionScreen.js
```diff
@@ -1,110 +1,79 @@
...
         </TextileHeaderButtons>
       ),
     }
   }
 
   handleNewText = (text: string) => {
-    this.setState({ value: text})
+    this.props.updateComment(text)
     this.props.navigation.setParams({
-      close: () => { this.props.close() },
```
asutula: This is cool, I discovered that `navigation.setParams` doesn't overwrite any params that were previously set, it just updates or adds to them. So you don't have to re-set the params that were already set in `componentWillMount`

App/Models/TextileTypes.ts
```diff
@@ -101,9 +101,7 @@ export type SharedImage = {
   path?: string,
   height: number,
   width: number,
-  isVertical: boolean,
-  caption?: string, // used for pre-upload display
```
asutula: Don't need `caption` and `addResult` in here because we track them next to `SharedImage` in `ProcessingImagesRedux` now

App/Navigation/AppNavigation/Primary/TabNavigator/Threads/index.ts
```diff
@@ -15,7 +15,7 @@ const nav = createStackNavigator(
     AddThread: AddThreadScreen,
     ViewThread: ThreadsDetail,
     PhotoViewer: ThreadPhotoDetail,
-    SharePhoto: AddCaptionScreen,
```
asutula: We had duplicate route names. Not sure what the expected behavior is in that case, but just made them unique.

App/Redux/AuthRedux.ts
```diff
@@ -117,7 +118,7 @@ export function reducer (state: AuthState = initialState, action: AuthAction): A
 }
 
 export const AuthSelectors = {
-  tokens: (state: any) => state.auth.tokens
+  tokens: (state: RootState) => state.auth.tokens
```
asutula: Can use `RootState` now wherever you have to handle state, nice and typed

App/Redux/CameraRollRedux.ts
```diff
@@ -21,24 +21,6 @@ const actions = {
...
-    return (threadId: string, image: TextileTypes.SharedImage, addResult: TextileTypes.AddResult) => resolve({ threadId, image, addResult })
-  }),
-  remotePinStarted: createAction('REMOTE_PIN_STARTED', resolve => {
-    return (threadId: string, image: TextileTypes.SharedImage) => resolve({ threadId, image })
-  }),
-  imagePinError: createAction('IMAGE_PICKER_PIN_ERROR', resolve => {
-    return (threadId: string, image: TextileTypes.SharedImage) => resolve({ threadId, image })
-  }),
-  cancelShare: createAction('CANCEL_IMAGE_PICKER_SHARE', resolve => {
-    return (threadId: string, image: TextileTypes.SharedImage) => resolve({ threadId, image })
```
asutula: It was confusing to have image sharing redux spread around, so it's all in `UIRedux` now

App/Redux/ProcessingImagesRedux.ts
```diff
@@ -0,0 +1,207 @@
...
+    readonly uploadProgress: number
+    readonly responseCode?: string
+    readonly responseBody?: string
+    readonly errorMessage?: string
+  }
+}
+
+export type ProcessingImagesState = {
+  readonly images: ProcessingImage[]
+}
```
asutula: This redux is where we track and persist the state of images from when they are first picked to be shared until they are done being shared to a thread. We track state though all of adding to IPFS, upload for remote pinning, adding to wallet (`default` thread), and sharing to thread.

App/Redux/UIRedux.ts
```diff
@@ -86,10 +87,9 @@ export type UIState = {
     readonly photoId?: string
     readonly threadId?: string
   },
-  readonly sharingPhoto: {
-    readonly active: boolean,
-    readonly hash?: string,
-    readonly selectedThreads: ReadonlyMap<string, boolean>,
+  readonly sharingPhoto?: {
```
asutula: Centralized the state of picking a photo to share, adding a comment, etc. Makes it easier to not have to pass state between screens in the nav props. Also, this state works for shares from the Wallet or from the camera roll (note `image?: SharedImage | string`).

App/Sagas/HandleImageUploadComplete.ts
```diff
@@ -0,0 +1,38 @@
+import { call, put, select } from 'redux-saga/effects'
```
asutula: Hey look, not adding any more to `TextileSagas`. Putting new sagas in separate files.

App/Sagas/HandleSharePhotoRequest.ts
```diff
@@ -0,0 +1,56 @@
...
+import RNFS from 'react-native-fs'
+
+import { uploadFile } from './UploadFile'
+import TextileNode from '../../TextileNode'
+import { SharedImage, AddResult } from '../Models/TextileTypes'
+import TextileNodeActions from '../Redux/TextileNodeRedux'
+import ProcessingImagesActions from '../Redux/ProcessingImagesRedux'
+import UIActions from '../Redux/UIRedux'
+
+export function * handleSharePhotoRequest(action: ActionType<typeof UIActions.sharePhotoRequest>) {
```
asutula: Step #1. This guy handles a share request by either doing the share from wallet thing, or by adding a photo from the camera roll to IPFS, and triggering an upload.

App/Sagas/HandleImageUploadComplete.ts
```diff
@@ -0,0 +1,38 @@
...
+import { ActionType } from 'typesafe-actions'
+import RNFS from 'react-native-fs'
+
+import TextileNode from '../../TextileNode'
+import ProcessingImagesActions, { ProcessingImagesSelectors, ProcessingImage } from '../Redux/ProcessingImagesRedux'
+import getDefaultThread from './GetDefaultThread'
+import { Thread } from '../Models/TextileTypes'
+
+export function * handleImageUploadComplete (action: ActionType<typeof ProcessingImagesActions.imageUploadComplete>) {
```
asutula: Step #2. Once the upload is complete, this guy adds the photo to the wallet and shared it to the thread the user originally chose.

App/Sagas/TextileSagas.ts
```diff
@@ -611,8 +576,9 @@ export function * showImagePicker(action: ActionType<typeof UIActions.showImageP
         width: pickerResponse.width,
         isVertical: pickerResponse.isVertical
       }
-      yield put(CameraRollActions.imagePickerSuccess(threadId, image))
-      yield call(NavigationService.navigate, 'SharePhoto', {threadId, image})
+      yield put(UIActions.updateSharingPhotoThread(threadId))
+      yield put(UIActions.updateSharingPhotoImage(image))
+      yield call(NavigationService.navigate, 'ThreadSharePhoto', { backTo: 'ViewThread' })
```
asutula: The AddComment screen can be canceled out of by tapping the back button, but depending on what context you're in (adding from the wallet screen or adding from the thread detail screen), you need to go back to a different place. To support that we pass in a route name as `backTo` that the AddComment screen can ready and use when the back button is tapped.

App/Models/TextileTypes.ts
andrewxhill: nice

App/Navigation/AppNavigation/Primary/TabNavigator/Threads/index.ts
andrewxhill: cool

App/Redux/AuthRedux.ts
andrewxhill: nice

App/Redux/CameraRollRedux.ts
andrewxhill: perfect

App/Redux/ProcessingImagesRedux.ts
andrewxhill: nice

App/Containers/AddCaptionScreen.js
```diff
@@ -1,110 +1,79 @@
...
   static navigationOptions = ({ navigation }) => {
     const params = navigation.state.params || {}
     return {
       headerTitle: 'Photo Caption',
       headerLeft: (
         <TextileHeaderButtons left>
-          <Item title='Back' iconName='arrow-left' onPress={() => { navigation.dispatch(NavigationActions.back()) }} />
+          <Item title='Back' iconName='arrow-left' onPress={() => {
+            params.cancelShare()
+            navigation.dispatch(NavigationActions.navigate({ routeName: params.backTo }))
```
asutula: This is cool. With the new version of `react-navigation` we updated to recently, you can just `navigate` to a route name and not worry about if that route is already in the stack. If it's not, it will be pushed onto the stack. If it is in the stack, it will navigate 'back' to it.

App/Redux/UIRedux.ts
andrewxhill: very cool

App/Sagas/GetDefaultThread.ts
```diff
@@ -0,0 +1,22 @@
+import { call, put, take } from 'redux-saga/effects'
+import { ActionType, getType } from 'typesafe-actions'
+
+import TextileNode from '../../TextileNode'
+import { Threads } from '../Models/TextileTypes'
+import ThreadsActions from '../Redux/ThreadsRedux'
+
+export default function * getDefaultThread () {
```
andrewxhill: nice

App/Sagas/HandleImageUploadComplete.ts
andrewxhill: yes! about time

App/Sagas/HandleSharePhotoRequest.ts
```diff
@@ -0,0 +1,56 @@
...
+    yield put(TextileNodeActions.getPhotoHashesRequest(threadId))
+  } catch (error) {
+    yield put(UIActions.imageSharingError(error))
+  }
+}
+
+function * addAndUploadImage (image: SharedImage, threadId: string, comment?: string) {
+  try {
+    yield put(ProcessingImagesActions.insertAddingImage(image, threadId, comment))
+    const path = image.uri.replace('file://', '')
```
andrewxhill: didn't notice, is the Android path variation being handled way back somewhere now?

asutula: Ah not sure. What does the android `uri` look like, do you know?

andrewxhill: the picker payload includes a `.path` you should use instead of `uri` i think

andrewxhill:
clooooose?!?

PR #382, opened 2018-08-17
Notifications

andrewxhill:


App/Containers/ThreadsList.js
```diff
@@ -56,24 +58,59 @@ class ThreadsList extends React.PureComponent {
...
       this.props.navigation.setParams({
-        profile: this.props.profile
+        profile: this.props.profile,
+        onTour: this.props.tourScreen === true
       })
     }
+    if (
+      this.props.threads.length !== prevProps.threads.length
+    ) {
+      this._notificationPrompt()
```
asutula: There's a really nice way to do things like this in a saga. Detaching it from the UI makes it a lot cleaner. Check out the `watchFirstThreeTodosCreation` saga here to see the idea: https://redux-saga.js.org/docs/advanced/FutureActions.html

App/Redux/PreferencesRedux.ts
```diff
@@ -46,7 +55,44 @@ export const initialState: PreferencesState = {
   verboseUi: false,
   tourScreens: {
     wallet: true,
-    threads: true
+    threads: true,
+    notifications: true
+  },
+  services: {
```
asutula: Cool, this works, thought just popped into my head that i wonder if it would be easier to deal with these as a tree instead of using `dependsOn`. So you'd have a root service (that is always enabled), it's children would be, for example, `notifications`, the the children of `notifications` would be all the different notification types, etc. Just a thought, if this is easy and scaleable, cool.

App/Redux/PreferencesRedux.ts
```diff
@@ -78,7 +129,8 @@ export const PreferencesSelectors = {
   // TODO: Need typed state
   onboarded: (state: any) => state.preferences.onboarded,
   pending: (state: any) => state.preferences.pending,
-  profile: (state: any) => state.preferences.profile
+  profile: (state: any) => state.preferences.profile,
+  service: (state: any, name: ServiceType) => state.preferences.services[name]
```
asutula: You can use `RootState` here and get all the TS goodness.

App/Redux/PreferencesRedux.ts
andrewxhill: pondered this one a bit more and realized, those nestings are primarily UI related, I'm going to remove them from the persisted state entirely

App/Redux/PreferencesRedux.ts
andrewxhill: done

App/Redux/PreferencesRedux.ts
andrewxhill: done

App/Redux/TextileNodeRedux.ts
```diff
@@ -146,15 +166,16 @@ function createEmptyThreadData (): ThreadData {
 
 // TODO: create RootState type that will be passed into these
 export const TextileNodeSelectors = {
-  locked: (state: any) => (state.textileNode as TextileNodeState).locked,
-  appState: (state: any) => (state.textileNode as TextileNodeState).appState,
-  online: (state: any) => (state.textileNode as TextileNodeState).online,
-  threads: (state: any) => (state.textileNode as TextileNodeState).threads,
+  locked: (state: RootState) => (state.textileNode as TextileNodeState).locked,
```
asutula: You can remove the type casts here now.

App/SB/views/Settings/index.js
```diff
@@ -63,60 +60,51 @@ class SyncPermissions extends React.PureComponent {
...
-                  </TouchableOpacity>
-                </View>
-              </View>
-              <Button title={'enable'} color={'#4a4a4a'} buttonStyle={styles.permissionsButtonStyle} onPress={this.toggleBackground.bind(this)} />
-            </View>
+            {Object.keys(this.props.services)
+              .map((service, i) => {
+                const value = !!this.props.services[service].status
+                let children = Object.keys(this.props.children)
+                  .filter((key) => this.props.children[key].info.dependsOn === service)
```
asutula: `dependsOn` is gone now, right?

App/SB/views/Settings/index.js
```diff
@@ -126,15 +114,43 @@ class SyncPermissions extends React.PureComponent {
...
+  const allServices: {[index: string]: string} = Object.keys(state.preferences.services)
+    .reduce((previous, current) => {
+      let basic = state.preferences.services[current]
+      basic.info = GetServiceInfo(current)
+      previous[current] = basic
+      return previous
+    }, {})
+
+  const services: {[index: string]: string} = Object.keys(allServices)
+    .filter((key) => allServices[key].info !== undefined && allServices[key].info.dependsOn === undefined)
```
asutula: `dependsOn` is used here too

App/SB/views/Settings/index.js
```diff
@@ -126,15 +114,43 @@ class SyncPermissions extends React.PureComponent {
...
+    .reduce((previous, current) => {
+      previous[current] = allServices[current]
+      return previous
+    }, {})
+
+  // get any services that depend on top level services
+  const children = Object.keys(allServices)
+    .filter((key) => allServices[key].info !== undefined && allServices[key].info.dependsOn !== undefined)
```
asutula: And here too

andrewxhill: ditto

App/SB/views/Settings/index.js
andrewxhill: I moved it into the `.info` object which is only here in the UI

App/SB/views/Settings/index.js
asutula: Oh I guess you removed it from the `ReduxPersist` migration, but it's still in `TextileTypes`. Not sure what your intention is there, let me know.

andrewxhill: nah, not in TTypes, i just append it here in this view. Adds an info object per key. Info contains what the title string, info string, nesting info are. All things just for the display. It's here, https://github.com/textileio/textile-mobile/pull/382/files#diff-ca34c21d7c542bf55f24da0b6f885ba9R1

asutula: Ayyy, just ignore me for a minute, think i'm all confused about where different things are defined/used.

App/Models/TextileTypes.ts
```diff
@@ -3,6 +3,38 @@ export interface Event {
...
+  peerLeftNotification
+}
+
+export interface Notification {
+  id: string
+  date: string
+  actor_id: string
+  actor_username: string
+  target_id: string
+  type: number
```
asutula: Does this correspond to a `NotificationType`? If so, we should use the actual type here.

App/Services/EventHandlers/NotificationEventHandler.ts
```diff
@@ -0,0 +1,38 @@
...
+    this.store = store
+    this.setup()
+  }
+
+  onNotification (notification: TextileTypes.NotificationEngagement) {
+    if (notification.userInteraction === true) {
+      this.store.dispatch(TriggersActions.newEngagement(notification))
+    }
+    if (notification.finish) {
+      notification.finish(PushNotificationIOS.FetchResult.NoData)
```
asutula: Not sure what this does, but just flagging as Android TODO since it's using an iOS only API

App/Services/Notifications.ts
```diff
@@ -0,0 +1,129 @@
...
+  if (!actor || actor === '') {
+    actor = 'Someone'
+  }
+  return actor
+}
+
+function notificationToPayload(notification: TextileTypes.Notification): NotificationsPayload {
+  const category = TextileTypes.NotificationType[notification.type] as string
+  switch (notification.type) {
+    case(0): {
```
asutula: Yea, think these should by typed enums if you update `TextileTypes` to use the enum instead of `number`.

android/app/build.gradle
```diff
@@ -166,6 +166,9 @@ android {
 }
 
 dependencies {
+    compile project(':react-native-extra-dimensions-android')
+    compile project(':react-native-push-notification')
+    compile project(':react-native-notifications')
```
asutula: No longer used, right?

asutula:
Looks like all traces of the wix react-native-notifications lib need to be removed from the Android and iOS projects

android/app/build.gradle
andrewxhill: cool yeah, haven't gone through android stuff yet

andrewxhill:
yep, haven't even tried on android

andrewxhill:
Okay, i think this guy is good for a whirl on TF.

There is going to be some work to make notifications really nice, but let's see if they even work. 

Feels like the very next step would be to get the background location enabled again, enough to warm up node and update notifications

App/Containers/ThreadsList.js
andrewxhill: opened a ticket for this, will handle in a follow-up.

App/Config/ReduxPersist.ts
```diff
@@ -103,14 +103,58 @@ const migrations: MigrationManifest = {
...
   }
 }
 
 const persistConfig: PersistConfig = {
   key: 'primary',
   storage: AsyncStorage,
-  version: 7,
-  whitelist: ['preferences', 'uploadingImages', 'processingImages', 'cameraRoll', 'notifications'],
+  version: 8,
+  whitelist: ['preferences', 'uploadingImages', 'processingImages', 'cameraRoll'],
```
asutula: So we don't want to persist notification data?

andrewxhill: nope! all comes from go now

App/Redux/NotificationsRedux.ts
```diff
@@ -1,36 +1,69 @@
...
+  notificationFailure: createAction('NOTIFICATION_FAILURE', resolve => {
+    return (notification: TextileTypes.Notification) => resolve({ notification })
   })
 }
 
 export type NotificationsAction = ActionType<typeof actions>
 
 export type NotificationsState = {
-  readonly notifications: Array<TextileTypes.Notification>
+  readonly notifications: Array<TextileTypes.Notification>,
```
asutula: Use `ReadonlyArray` here.

App/Redux/PreferencesRedux.ts
```diff
@@ -24,20 +25,28 @@ const actions = {
...
+}
 export type PreferencesState = {
   onboarded: boolean
   verboseUi: boolean
   mnemonic?: string
   publicKey?: string
   profile?: Profile
   pending?: string,
-  // tracks if a user has seen a particular tour screen
+  readonly services: {[index: string]: Service}
```
asutula: Maybe this should be indexed by `ServiceType` and not `string`?

andrewxhill: you cannot have both in the map like that

andrewxhill: as in, `{[index: ServiceType]: Service}` isn't valid

asutula: Hmm. Maybe if you make the enum?

App/Redux/UIRedux.ts
```diff
@@ -63,14 +64,8 @@ const actions = {
...
   }),
-  refreshMessagesRequest: createAction('REFRESH_MESSAGES_REQUEST', resolve => {
-    return (hidden?: boolean) => resolve({hidden})
-  }),
-  refreshMessagesSuccess: createAction('REFRESH_MESSAGES_SUCCESS', resolve => {
-    return (timestamp: number) => resolve({ timestamp })
-  }),
-  refreshMessagesFailure: createAction('REFRESH_MESSAGES_FAILURE', resolve => {
-    return (error: Error) => resolve({ error })
+  routeDeepLinkRequest: createAction('ROUTE_DEEP_LINK_REQUEST', resolve => {
```
asutula: This is duplicated here and in `TriggersRedux`

App/Redux/PreferencesRedux.ts
andrewxhill: could be yeah, sounds like future optimization. by using the `ServiceType` for all action inputs where it occurs it's 99% typed :)

asutula: Sorry, i meant maybe if you make `ServiceType` an `enum`.

andrewxhill: ah yeah, could be

andrewxhill: just tried, must be string or number

andrewxhill: make all of services an enum i guess. seriously though, next time

andrewxhill: ha

App/Redux/UIRedux.ts
andrewxhill: doh, fixed. no idea why i did that

PR #390, opened 2018-08-21
Share retry/cancel

asutula:
Supports image sharing retry and cancel. UI looks bad, need to make some nice icons (#391).


App/Containers/AddCaptionScreen.js
```diff
@@ -20,7 +20,7 @@ class AddCaptionScreen extends React.Component {
           }} />
         </TextileHeaderButtons>
       ),
-      headerRight: params.submitEnabled && (
+      headerRight: (
```
asutula: Allow sharing with no caption in the UI. Redux layer already supports it. cc: @sanderpick

App/Redux/ProcessingImagesRedux.ts
```diff
@@ -3,48 +3,61 @@ import { SharedImage, AddResult } from '../Models/TextileTypes'
 import { RootState } from './Types'
 
 const actions = {
-  insertAddingImage: createAction('processingImages/INSERT_ADDING_IMAGE', resolve => {
-    return (sharedImage: SharedImage, destinationThreadId: string, comment?: string) => resolve({ sharedImage, destinationThreadId, comment })
+  insertImage: createAction('processingImages/INSERT_IMAGE', resolve => {
+    return (uuid: string, sharedImage: SharedImage, destinationThreadId: string, comment?: string) => resolve({ uuid, sharedImage, destinationThreadId, comment })
```
asutula: Decided to have processing images be tracked with a uuid. This simplified things because we need to track processing images before and after they are added to IPFS, and we don't get a unique id from IPFS until after the image is added to IPFS.

App/Containers/AddCaptionScreen.js
sanderpick: sweet. will it be safe for older versions if the api starts sending objects without a caption key?

App/Sagas/ImageSharingSagas.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+}
+
+export function * insertImage (image: SharedImage, threadId: string, comment?: string) {
+  const id = uuid()
+  yield put(ProcessingImagesActions.insertImage(id, image, threadId, comment))
+  yield call(addToIpfs, id)
+}
+
+export function * addToIpfs (uuid: string) {
```
asutula: Idea here is that all these sagas should be able to be called repeatedly (for retry), and the redux state will be consistent and the sequence of things that need to happen to share a photo will continue no matter where you start the retry from in the sequence.

App/Sagas/ImageSharingTriggers.ts
```diff
@@ -0,0 +1,56 @@
+import { call, put, select } from 'redux-saga/effects'
```
asutula: These are the functions that can kick off photo sharing actions.

App/Sagas/ImageSharingTriggers.ts
```diff
@@ -0,0 +1,56 @@
...
+import { ActionType } from 'typesafe-actions'
+import RNFS from 'react-native-fs'
+
+import ProcessingImagesActions, { ProcessingImage, ProcessingImagesSelectors } from '../Redux/ProcessingImagesRedux'
+import UIActions from '../Redux/UIRedux'
+import { insertImage, addToIpfs, uploadArchive, shareWalletImage, addToWallet, shareToThread } from './ImageSharingSagas'
+
+export function * handleSharePhotoRequest(action: ActionType<typeof UIActions.sharePhotoRequest>) {
```
asutula: This one handled the UI action of sharing a photo, either from the wallet or camera roll.

App/Sagas/ImageSharingTriggers.ts
```diff
@@ -0,0 +1,56 @@
...
+    return
+  }
+  if (typeof image === 'string') {
+    yield call(shareWalletImage, image, threadId, comment)
+  } else {
+    yield call(insertImage, image, threadId, comment)
+  }
+}
+
+export function * handleImageUploadComplete (action: ActionType<typeof ProcessingImagesActions.imageUploadComplete>) {
```
asutula: This one picks up the process when we're notified an image has finished uploading.

App/Sagas/ImageSharingTriggers.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (processingImage && processingImage.addData && processingImage.addData.addResult.archive) {
+      const exists: boolean = yield call(RNFS.exists, processingImage.addData.addResult.archive.path)
+      if (exists) {
+        yield call(RNFS.unlink, processingImage.addData.addResult.archive.path)
+      }
+    }
+  } catch (e) {}
+}
+
+export function * retryImageShare (action: ActionType<typeof ProcessingImagesActions.retry>) {
```
asutula: This one examines where in the process an error happened and retries from that point on.

App/Sagas/ImageSharingTriggers.ts
```diff
@@ -0,0 +1,56 @@
...
+  } else if (processingImage.state === 'uploaded' || processingImage.state === 'addingToWallet') {
+    yield call(addToWallet, uuid)
+  } else if (processingImage.state === 'added' || processingImage.state === 'uploading') {
+    yield call(uploadArchive, uuid)
+  } else if (processingImage.state === 'pending' || processingImage.state ==='adding') {
+    yield call(addToIpfs, uuid)
+  }
+}
+
+export function * cancelImageShare (action: ActionType<typeof ProcessingImagesActions.cancelRequest>) {
```
asutula: This cancels a share. Need to discuss what exactly need to be/can be done here.

App/Sagas/ImageSharingSagas.ts
andrewxhill: cool

App/Sagas/ImageSharingTriggers.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    // Delete the Textile payload
+    if (processingImage.addData && processingImage.addData.addResult.archive) {
+      const exists: boolean = yield call(RNFS.exists, processingImage.addData.addResult.archive.path)
+      if (exists) {
+        yield call(RNFS.unlink, processingImage.addData.addResult.archive.path)
+      }
+    }
+
+    // What else? Undo local add, remote pin, remove from wallet?
```
asutula: Ok, I added code to delete the image and upload archive. What else can we do here?

asutula: I don't display the cancel button unless the share has failed. Another option is to always display the cancel button to allow canceling of in-flight shares. That might complicate this more, not sure. Thoughts appreciated.

andrewxhill:
looks good, but still wip. lemme know

App/Models/TextileTypes.ts
```diff
@@ -42,7 +42,7 @@ export interface Photo {
   block_id: string
   date: string
   author_id: string
-  caption: string
+  caption?: string
```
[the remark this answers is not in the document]
sanderpick: cool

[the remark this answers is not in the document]
sanderpick: will add the `omitempty` flag to the api type

App/Sagas/ImageSharingTriggers.ts
asutula: @andrewxhill and @sanderpick this is still WIP mostly because of this question.

andrewxhill: +1 for the way you did it

asutula:
Made issue #391 for adding nice cancel and retry icons.

App/Sagas/ImageSharingTriggers.ts
sanderpick: - If it failed _after_ being added to IPFS, we should eventually unpin from local node (not all of it is pinned to begin with, but still would be good to unpin the thumb, small, and meta). I guess this would only happen if the add / share to thread failed. I think skippable for now... let's see what breaks first :)

asutula: ok cool.

asutula:
Ok this is G2G

PR #394, opened 2018-08-21
Make tour screen and service keys typesafe

asutula:
Did a little experiment just to see how this works. Might as well use it. Pretty cool actually.

### DIFF
--- a/App/Redux/PreferencesRedux.ts
+++ b/App/Redux/PreferencesRedux.ts
@@ -46,8 +46,8 @@ export type PreferencesState = {
   publicKey?: string
   profile?: Profile
   pending?: string,
-  readonly services: {[index: string]: Service}
-  readonly tourScreens: {[index: string]: boolean} // true = still need to show, false = no need
+  readonly services: {[k in ServiceType]: Service}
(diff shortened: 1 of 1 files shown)


App/Redux/PreferencesRedux.ts
```diff
@@ -46,8 +46,8 @@ export type PreferencesState = {
   publicKey?: string
   profile?: Profile
   pending?: string,
-  readonly services: {[index: string]: Service}
-  readonly tourScreens: {[index: string]: boolean} // true = still need to show, false = no need
+  readonly services: {[k in ServiceType]: Service}
+  readonly tourScreens: {[k in TourScreens]: boolean} // true = still need to show, false = no need
```
asutula: This enforces that only certain keys are used. Actually in this case it also ensures that **all** keys are used. If you wanted to ensure that the keys that are used are part of a known set, but you didn't want to require that all the keys are used, you could do:
```
readonly services: {[k in ServiceType]?: Service}
```

andrewxhill:
nice, swear i tried that but maybe pre some other thing... ha

PR #399, opened 2018-08-23
small fix on feed to show when empty

andrewxhill:
forgot again that the init state of arrays is len=0. avoids showing the placeholder after the user has used the app


andrewxhill:
added to other merge

PR #402, opened 2018-08-23
Node lifecycle, background triggers

asutula:
Feeling good about the node start/stop logic. Seems solid and will always run the node in the background for 20 seconds, whether the background session results from a foreground > background transition or from launching into the background because of a location or background fetch trigger.

Need some help:

I want to clean up all the things we do after the node is started and/or comes online. I deleted a bunch of old code, code that did some of those important things. Go through it with me and let's identify all the things that need to happen and exactly when/how they should happen.

### DIFF
--- a/App/Sagas/TextileSagas.ts
+++ b/App/Sagas/TextileSagas.ts
@@ -206,93 +206,6 @@ export function * initializeAppState () {
   yield put(TextileNodeActions.appStateChange(defaultAppState, queriedAppState))
 }
 
-export function * handleNewAppState (action: ActionType<typeof TextileNodeActions.appStateChange>) {
-  const { previousState, newState } = action.payload
-  if (previousState.match(/default|unknown/) && newState === 'background') {
-    yield * triggerCreateNode()
-  } else if (previousState.match(/default|unknown|inactive|background/) && newState === 'active') {
-    yield put(TextileNodeActions.lock(false))
-    yield * triggerCreateNode()
-  } else if (previousState.match(/inactive|active/) && newState === 'background') {
-    yield * triggerStopNode()
-  }
-}
-
-export function * triggerCreateNode () {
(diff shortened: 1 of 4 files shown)


asutula:
Need to get out of this cafe, but going to leave some comments in this code as soon as I'm home.

App/Sagas/TextileSagas.ts
```diff
@@ -206,93 +206,6 @@ export function * initializeAppState () {
...
-
-  try {
-    yield call(TextileNode.start)
-    yield put(TextileNodeActions.startNodeSuccess())
-
-    // Restore our tokens and username
-    const tokens = yield call(TextileNode.getTokens)
-    yield put(AuthActions.getTokensSuccess(tokens))
-
-    yield put(ThreadsActions.refreshThreadsRequest())
```
asutula: After node is started, we request that the threads be refreshed. Specifically, this gets the list of threads from the node and stores it in ThreadsRedux memory and triggers a fetch of the hashes for each thread and stores those in TextileNodeRedux memory.

App/Sagas/TextileSagas.ts
```diff
@@ -206,93 +206,6 @@ export function * initializeAppState () {
...
-
-    // isolate
-    try {
-      const publicKey = yield call(TextileNode.getPublicKey)
-      yield put(PreferencesActions.getPublicKeySuccess(publicKey))
```
asutula: After the node is started, we get the public key from the node and store it in PreferencesRedux persistence.

App/Sagas/TextileSagas.ts
sanderpick: You're probably doing this, but just making sure... can we call `refreshMessages` on the node too?

andrewxhill:
went through, looks good, let's give it a go!

PR #403, opened 2018-08-23
New notifications & very typed api responses

andrewxhill:


App/Components/Avatar.js
```diff
@@ -9,12 +9,18 @@ const Avatar = props => {
...
+        {uri && <Image
           source={{uri: uri}}
           resizeMode={'cover'}
           style={{width, height}}
           defaultSource={defaultSource}
-        />
+        />}
+        {/* avoids a null uri warning */}
+        {!uri && defaultSource && <Image
+          source={defaultSource}
```
andrewxhill: it's nice to be able to render the avatar style from a static icon sometimes (e.g. the feedview)

App/Models/TextileTypes.ts
```diff
@@ -1,19 +1,42 @@
...
+enum ProfileAvatarIdBase {}
+enum NotificationIdBase {}
+export type TextileId = TextileIdBase & string & undefined
+export type BlockId = BlockIdBase & string & undefined
+export type ThreadId = ThreadIdBase & string & undefined
+export type PhotoId = PhotoIdBase & string & undefined
+export type PeerId = PeerIdBase & string & undefined
+export type DeviceId = DeviceIdBase & string & undefined
+export type ProfileAvatarId = ProfileAvatarIdBase & string & undefined
+export type NotificationId = NotificationIdBase & string & undefined
```
andrewxhill: the enum base makes it so that these aren't simply strings that can be treated as strings, you'll actually get back an error if you try to pass, for example, a PeerId as a BlockId

andrewxhill: the undefined option is common for base types but here isn't, so when creating composite types that are like `preferences = {photoId?: TextileType.PhotoId}` you can't set it since PhotoId type is never undefined, but the photoId is looking for an either/or.

App/Redux/DevicesRedux.ts
```diff
@@ -53,16 +53,16 @@ export const initialState: DevicesState = {
 export function reducer (state: DevicesState = initialState, action: DevicesAction): DevicesState {
   switch (action.type) {
     case getType(actions.addDeviceRequest): {
-      const { name, pubKey } = action.payload
-      const deviceItem: TextileTypes.Device = {name, id: pubKey}
+      const { name, deviceId } = action.payload
+      const deviceItem: TT.Device = {name, id: deviceId}
```
andrewxhill: deviceid is the pubkey currently, so more clear here to just call it such

App/Redux/ProcessingImagesRedux.ts
```diff
@@ -54,24 +54,24 @@ export type ProcessingImagesAction = ActionType<typeof actions>
 
 export type ProcessingImage = {
   readonly uuid: string,
-  readonly sharedImage: SharedImage
-  readonly destinationThreadId: string
+  readonly sharedImage: TT.SharedImage
+  readonly destinationThreadId: TT.ThreadId
```
andrewxhill: Got sick of typing out TextileTypes all over, so standardized to TT

App/Containers/TextilePhotos.js
```diff
@@ -1,22 +1,18 @@
...
-import ActionSheet from 'react-native-actionsheet'
+import {View, Text, Image } from 'react-native'
 import HeaderButtons, { Item } from 'react-navigation-header-buttons'
 import PhotoGrid from '../Components/PhotoGrid'
 import { connect } from 'react-redux'
 import Config from 'react-native-config'
 import PreferencesActions from '../Redux/PreferencesRedux'
 import TextileNodeActions, { ThreadData } from '../Redux/TextileNodeRedux'
-import * as TextileTypes from '../Models/TextileTypes'
+import * as TT from '../Models/TextileTypes'
```
asutula: Would probably just be better to import the types you actually use, then the module alias isn't used at all.

App/Redux/CameraRollRedux.ts
```diff
@@ -43,8 +41,8 @@ export const initialState: CameraRollState = {
 }
 
 export const cameraRollSelectors = {
-  initialized: state => state.cameraRoll.initialized as boolean,
-  queriedPhotos: state => state.cameraRoll.queriedPhotos as QueriedPhotosMap
+  initialized: (state: RootState) => state.cameraRoll.initialized as boolean,
+  queriedPhotos: (state: RootState) => state.cameraRoll.queriedPhotos as QueriedPhotosMap
```
asutula: No casts needed here now

App/Redux/ProcessingImagesRedux.ts
asutula: See my comment in `TextilePhotos`.

App/Sagas/ImageSharingSagas.ts
```diff
@@ -14,14 +14,14 @@ import {ActionType} from 'typesafe-actions'
 export function * shareWalletImage (id: string, threadId: string, comment?: string) {
   try {
     // TODO: Insert some state into the processing photos redux in case this takes long or fails
-    const blockId: string = yield call(TextileNode.sharePhotoToThread, id, threadId, comment)
+    const blockId: TT.BlockId = yield call(TextileNode.sharePhotoToThread, id, threadId, comment)
     yield put(TextileNodeActions.getPhotoHashesRequest(threadId))
```
asutula: There are still places like this where you're passing a `string` (`threadId`) into something typed with the enum (`getPhotoHashesRequest`). Do we want to go though and make sure we take the typing as far as we can? I'm guessing the string -> enum un-typed scary conversion could be isolated to API responses and input from the user (the UI).

asutula: Looks like other similar cases are typed, so maybe I just happened to look at the one you forgot...

andrewxhill: yeah, shouldn't ever be typed as string anymore

App/Redux/CameraRollRedux.ts
asutula: But now that it's typed as`RootState` they type shouldn't be `any`.

andrewxhill: ah, sorry, mistook what you were pointing at

App/Sagas/TextileSagas.ts
```diff
@@ -32,11 +32,10 @@ import UIActions from '../Redux/UIRedux'
 import ThreadsActions from '../Redux/ThreadsRedux'
 import DevicesActions from '../Redux/DevicesRedux'
 import { ActionType, getType } from 'typesafe-actions'
-import * as TextileTypes from '../Models/TextileTypes'
+import * as TT from '../Models/TextileTypes'
```
andrewxhill: leaves it here

TextileNode.ts
```diff
@@ -4,8 +4,8 @@ import {
   NativeEventEmitter,
   DeviceEventEmitter,
 } from 'react-native'
-import * as TextileTypes from './App/Models/TextileTypes'
-import {Notification} from './App/Models/TextileTypes'
+import * as TT from './App/Models/TextileTypes'
```
andrewxhill: and here. otherwise just grabs the exact ones

andrewxhill:
should be gtg

PR #405, opened 2018-08-24
Node lifecycle, background triggers (2)

andrewxhill:
Feeling good about the node start/stop logic. Seems solid and will always run the node in the background for 20 seconds, whether the background session results from a foreground > background transition or from launching into the background because of a location or background fetch trigger.

Need some help:

I want to clean up all the things we do after the node is started and/or comes online. I deleted a bunch of old code, code that did some of those important things. Go through it with me and let's identify all the things that need to happen and exactly when/how they should happen.

Todo:
- [x] Fire off things that should happen once the node is created
- [x] Fire off things that should happen once the node is started
- [x] Create UI and logic for enabling background location

### DIFF
--- a/App/Redux/TextileNodeRedux.ts
+++ b/App/Redux/TextileNodeRedux.ts
@@ -4,9 +4,6 @@ import * as TextileTypes from '../Models/TextileTypes'
 import {RootState} from './Types'
 
 const actions = {
-  lock: createAction('LOCK', resolve => {
-    return (value: boolean) => resolve({ value })
-  }),
   appStateChange: createAction('APP_STATE_CHANGE', resolve => {
     return (previousState: TextileAppStateStatus, newState: AppStateStatus) => resolve({ previousState, newState })
   }),
@@ -72,7 +69,7 @@ type TextileAppStateStatus = AppStateStatus | 'unknown'
 export enum NodeState {
   'nonexistent' = 'nonexistent',
   'creating' = 'creating',
-  'created' = 'created', // Node has been created, on it's way to being starting
+  'created' = 'created', // Node has been created, on it's way to starting
   'starting' = 'starting',
   'started' = 'started',
   'stopping' = 'stopping',
@@ -80,7 +77,6 @@ export enum NodeState {
 }
 
 type TextileNodeState = {
-  readonly locked: boolean
   readonly appState: TextileAppStateStatus
   readonly online: boolean
   readonly nodeState: {
@@ -92,7 +88,6 @@ type TextileNodeState = {
 }
 
 export const initialState: TextileNodeState = {
-  locked: false,
   appState: 'unknown',
   online: false,
   nodeState: {
@@ -104,8 +99,6 @@ export const initialState: TextileNodeState = {
 
 export function reducer (state: TextileNodeState = initialState, action: TextileNodeAction): TextileNodeState {
   switch (action.type) {
-    case getType(actions.lock):
-      return { ...state, locked: action.payload.value }
     case getType(actions.appStateChange):
       return { ...state, appState: action.payload.newState }
     case getType(actions.creatingNode):
--- a/App/Redux/TextileNodeRedux.ts
+++ b/App/Redux/TextileNodeRedux.ts
@@ -10,32 +10,18 @@ const actions = {
   appStateChange: createAction('APP_STATE_CHANGE', resolve => {
     return (previousState: TextileAppStateStatus, newState: AppStateStatus) => resolve({ previousState, newState })
   }),
-  createNodeRequest: createAction('CREATE_NODE_REQUEST', resolve => {
-    return (path: string) => resolve({ path })
-  }),
-  createNodeSuccess: createAction('CREATE_NODE_SUCCESS', resolve => {
-    return () => resolve()
-  }),
-  createNodeFailure: createAction('CREATE_NODE_FAILURE', resolve => {
-    return (error: Error) => resolve({ error })
-  }),
-  startNodeRequest: createAction('START_NODE_REQUEST', resolve => {
-    return () => resolve()
-  }),
+  creatingNode: createAction('CREATING_NODE'),
+  createNodeSuccess: createAction('CREATE_NODE_SUCCESS'),
+  startingNode: createAction('STARTING_NODE'),
   startNodeSuccess: createAction('START_NODE_SUCCESS', resolve => {
     return () => resolve()
   }),
-  startNodeFailure: createAction('START_NODE_FAILURE', resolve => {
-    return (error: Error) => resolve({ error })
-  }),
-  stopNodeRequest: createAction('STOP_NODE_REQUEST', resolve => {
-    return () => resolve()
-  }),
+  stoppingNode: createAction('STOP_NODE_REQUEST'),
   stopNodeSuccess: createAction('STOP_NODE_SUCCESS', resolve => {
     return () => resolve()
   }),
-  stopNodeFailure: createAction('STOP_NODE_FAILURE', resolve => {
-    return (error: Error) => resolve({ error })
+  nodeError: createAction('NODE_ERROR', resolve => {
+    return (error: any) => resolve({ error })
   }),
   nodeOnline: createAction('NODE_ONLINE', resolve => {
     return () => resolve()
@@ -83,13 +69,23 @@ type ThreadMap = {
 
 type TextileAppStateStatus = AppStateStatus | 'unknown'
 
+export enum NodeState {
+  'nonexistent' = 'nonexistent',
+  'creating' = 'creating',
+  'created' = 'created', // Node has been created, on it's way to being starting
+  'starting' = 'starting',
+  'started' = 'started',
+  'stopping' = 'stopping',
+  'stopped' = 'stopped' // Node has been explicitly stopped, different than created
+}
+
 type TextileNodeState = {
   readonly locked: boolean
   readonly appState: TextileAppStateStatus
   readonly online: boolean
   readonly nodeState: {
-    readonly state?: 'creating' | 'stopped' | 'starting' | 'started' | 'stopping'
-    readonly error?: Error
+    readonly state: NodeState
+    readonly error?: string
   }
   readonly threads: ThreadMap,
   readonly refreshingMessages: boolean
@@ -99,7 +95,9 @@ export const initialState: TextileNodeState = {
   locked: false,
   appState: 'unknown',
   online: false,
-  nodeState: {},
+  nodeState: {
+    state: NodeState.nonexistent
+  },
   threads: {},
   refreshingMessages: false
 }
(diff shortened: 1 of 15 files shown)


App/Sagas/TextileSagas.ts
```diff
@@ -206,93 +197,6 @@ export function * initializeAppState () {
...
-export function * createNode (action: ActionType<typeof TextileNodeActions.createNodeRequest>) {
-  const { path } = action.payload
-  try {
-    const logLevel = (__DEV__ ? 'DEBUG' : 'INFO')
-    const logFiles = !__DEV__
-    yield call(TextileNode.create, path, Config.TEXTILE_CAFE_URI, logLevel, logFiles)
-    yield put(TextileNodeActions.createNodeSuccess())
-    try {
-      const mnemonic: string = yield call(TextileNode.mnemonic)
-      yield put(PreferencesActions.updatecMnemonic(mnemonic))
```
asutula: After node is created, we updated preferences with the mnemonic from the node.

App/Sagas/TextileSagas.ts
```diff
@@ -206,93 +197,6 @@ export function * initializeAppState () {
...
-    return
-  }
-
-  try {
-    yield call(TextileNode.start)
-    yield put(TextileNodeActions.startNodeSuccess())
-
-    // Restore our tokens and username
-    const tokens = yield call(TextileNode.getTokens)
-    yield put(AuthActions.getTokensSuccess(tokens))
```
asutula: After node is started, we get auth tokens and store them in memory

App/Sagas/TextileSagas.ts
```diff
@@ -206,93 +197,6 @@ export function * initializeAppState () {
...
-
-    yield put(ThreadsActions.refreshThreadsRequest())
```
asutula: After node is started, we request that the threads be refreshed. Specifically, this gets the list of threads from the node and stores it in ThreadsRedux memory and triggers a fetch of the hashes for each thread and stores those in TextileNodeRedux memory.

asutula: @sanderpick replied:

> You're probably doing this, but just making sure... can we call refreshMessages on the node too?

asutula: No, at this point I'm not doing anything... I want to make sure I get a list of all the things that need to be done and exactly when they should be done. Cool, yes, we can refresh messages in the background.

App/Sagas/TextileSagas.ts
```diff
@@ -206,93 +197,6 @@ export function * initializeAppState () {
...
-
-    // isolate
-    try {
-      const publicKey = yield call(TextileNode.getPublicKey)
-      yield put(PreferencesActions.getPublicKeySuccess(publicKey))
```
asutula: After the node is started, we get the public key from the node and store it in PreferencesRedux persistence.

sanderpick:
@asutula just catching up here and thinking what needs doing in BG. Maybe useful to start with that the goals are of launching in the BG (maybe incorrect, just exploring)...
1. To get the latest "state" in RN so that when the user opens the app, it's already up-to-date
2. To create any new local notifications based on new notifications
For 1., I _think_ this can be achieved by just calling `refreshingMessages`, nothing else:
- Threads won't be updated unless a user creates one, or accepts an invite (so, in-app stuff)
- All other new stuff (photos, notifications, comments, likes) will get pushed to the RN layer if the message retriever finds anything... other types of updates not visible to the UI will also be triggered (joined, leaves, merges, ignores)

So, a manual call to update the thread list and photo lists doesn't seem useful... maybe redundant if say a new photo is found, pushed to RN, which then refreshes that thread again.

sanderpick:
And then for 2., same deal... new notifications would have to be driven by successfully finding offline messages.

asutula:
@sanderpick does that call to `refreshMessages` need to happen once the node has started or once it's online?

sanderpick:
Online

App/Sagas/NodeCreated.ts
```diff
@@ -0,0 +1,17 @@
+import { call, put, take } from 'redux-saga/effects'
+import { getType } from 'typesafe-actions'
+import TextileNode from '../../TextileNode'
+import PreferencesActions from '../Redux/PreferencesRedux'
+import TextileNodeActions from '../Redux/TextileNodeRedux'
+
+
+export function * onNodeCreated () {
```
asutula: Wire up all things that should happen once the node is created

App/Sagas/NodeOnline.ts
```diff
@@ -0,0 +1,9 @@
+import { put, take } from 'redux-saga/effects'
+import { getType } from 'typesafe-actions'
+import TextileNodeActions from '../Redux/TextileNodeRedux'
+
+export function * onNodeOnline () {
```
asutula: Wire up all things that should happen once the node is online. This is where refreshing messages trigger happens @sanderpick

App/Sagas/NodeStarted.ts
```diff
@@ -0,0 +1,27 @@
+import { all, call, put, take } from 'redux-saga/effects'
+import { getType } from 'typesafe-actions'
+import TextileNode from '../../TextileNode'
+import AuthActions from '../Redux/AuthRedux'
+import PreferencesActions from '../Redux/PreferencesRedux'
+import TextileNodeActions from '../Redux/TextileNodeRedux'
+import ThreadsActions from '../Redux/ThreadsRedux'
+
+export function * onNodeStarted () {
```
asutula: Wire up all things that happen once the node is started. 

Possible optimization: Split this into two functions, one for node started in background and one for node started in foreground. That would allow us to only do UI related things in the node has started and we're in the foreground.

App/Sagas/TextileSagas.ts
```diff
@@ -149,7 +149,7 @@ function * processAvatarImage(uri: string, defaultThread: TT.Thread) {
 }
 
 export function * viewThread ( action: ActionType<typeof UIActions.viewThreadRequest> ) {
-  yield call(TextileNodeActions.refreshMessagesRequest, false)
+  yield put(TextileNodeActions.refreshMessagesRequest())
```
asutula: Think this was a bug @andrewxhill , so I fixed it. The old version would do nothing.

App/Sagas/TextileSagas.ts
asutula: Moved to node lifecycle listener.

App/Sagas/TextileSagas.ts
asutula: Moved to node lifecycle listener.

App/Sagas/TextileSagas.ts
asutula: Moved to node lifecycle listener.

App/Sagas/NodeStarted.ts
carsonfarmer: Separating them out makes sense to me @asutula. Certainly keeps things cleaner?

App/Sagas/index.ts
```diff
@@ -16,6 +16,12 @@ import DevicesActions from '../Redux/DevicesRedux'
 /* ------------- Sagas ------------- */
 
 import { startup } from './StartupSagas'
+
+import { manageNode } from './NodeLifecycle'
+import { onNodeCreated } from './NodeCreated'
+import { onNodeStarted } from './NodeStarted'
+import { onNodeOnline } from './NodeOnline'
```
asutula: Wire up the node lifecycle listeners

App/Sagas/index.ts
```diff
@@ -81,6 +81,11 @@ import {
 
 export default function * root () {
   yield all([
+    call(manageNode),
+    call(onNodeCreated),
+    call(onNodeStarted),
+    call(onNodeOnline),
```
asutula: Wire up the node lifecycle listeners

asutula: Becoming a fan of not using `takeEvery` etc in the root saga. It's cleaner to just yield a `call()` and let the called function contain the `take` logic is wants. More self contained that way, and the root saga is more like a manifest with no logic.

App/Sagas/NodeOnline.ts
sanderpick: Ok, cool. To clarify, I think we should call refresh messages when:
- node comes online
- pull down in any of these view: threads, a single thread, notifications

Not sure we should call it when you just view the above views (not pull down). 

In any case, I need to ensure we're not firing off a bunch of these requests in the API layer.

App/Redux/TextileNodeRedux.ts
```diff
@@ -147,7 +135,7 @@ export function reducer (state: TextileNodeState = initialState, action: Textile
       const threads = { ...state.threads, [threadId]: { ...threadData, querying: false, error } }
       return { ...state, threads }
     }
-    case getType(actions.refreshMessagesSubmitted):
+    case getType(actions.refreshMessagesRequest):
```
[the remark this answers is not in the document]
andrewxhill: No, sorry, this is going the wrong way. Every time there is a Request, you'd toggle refreshing to true. Then in TextileSages, when you go to check if you are already refreshing, you'd always find it to true. I must have somehow dropped the edit, but ~L300 of TextileSagas after `if (!refreshing) {` you actually want to call `TextileNodeActions.refreshMessagesSubmitted` which would only now toggle refreshing to true. Make sense? 

I think it got dropped along the way (100% my fault i'm guessing), I've noticed refresh not working very well since the original branch, now it makes sense

App/SB/views/ThreadsDetail/index.js
```diff
@@ -272,7 +272,7 @@ const mapDispatchToProps = (dispatch) => {
     dismissPhoto: () => { dispatch(UIActions.dismissViewedPhoto()) },
     viewPhoto: (photoId, threadId) => { dispatch(UIActions.viewPhotoRequest(photoId, threadId)) },
     showImagePicker: (threadId) => { dispatch(UIActions.showImagePicker(threadId)) },
-    refreshMessages: (hidden) => { dispatch(TextileNodeActions.refreshMessagesRequest(hidden)) },
```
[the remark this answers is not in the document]
andrewxhill: yeah, drop it. wild, there is def a clean up commit not included here. the idea was to avoid showing the spinner wheel on refresh during screen load. but kept it simple to start by just removing it

App/Sagas/NodeOnline.ts
asutula: K, so with this node online listener, messages are refreshed. We can go through the UI related code and see where else it's being called and make sure it's only in pull to refresh cases.

App/Redux/TextileNodeRedux.ts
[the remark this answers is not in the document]
asutula: Ok, I think this is something we can just solve by constructing the saga in a better way. `takeEvery` and `takeLatest` aren't good for when you want to respond to actions in a serial manner. Improved version of this on the way...

asutula:
Ok give this a spin. Think it's G2G.

asutula:
Created #407, #408 and #409 for some things I noticed along the way, but don't need to block us for now.

App/Sagas/NodeStarted.ts
asutula: Will do a a future revision

PR #410, opened 2018-08-25
removes push notification handlers in appdelegate

andrewxhill:


ios/TextilePhotos.xcodeproj/project.pbxproj
```diff
@@ -1205,6 +1205,9 @@
 							com.apple.BackgroundModes = {
 								enabled = true;
 							};
+							com.apple.Push = {
+								enabled = 0;
+							};
```
andrewxhill: Check that out. Automatically  updated by removing the methods from AppDelegate

andrewxhill:
let's see what apple has to say

ios/TextilePhotos.xcodeproj/project.pbxproj
asutula: wooooow

PR #411, opened 2018-08-25
Move location monitor

asutula:
Issue is we need access to the persisted preferences state to decide whether or not to start monitoring. That state wasn't available in `App.tsx` (too early, state not rehydrated from persistence yet). So for now, I moved the logic to `RootContainer` which isn't rendered until persistence is done initializing. Not the long term solution... I think we can do this as a saga real nice, just need a little more time to figure it out not on a Friday night.


asutula:
Merging lol

PR #426, opened 2018-08-28
Updated wallet view

andrewxhill:


App/Components/Styles/WalletHeader.tsx
```diff
@@ -0,0 +1,66 @@
...
+    borderWidth: 2,
+    borderRadius: 2,
+    alignContent: 'center',
+    justifyContent: 'center'
+  },
+  walletSettingsText: {
+    color: 'rgba(20,20,33,0.75)',
+    textAlign: 'center'
+  }
+})
```
asutula: Cool for now, but we should explore best practices for styling in TS. There are TS types for all the different types of styles (view style, font style, etc), and would be good and helpful to take advantage of those.

App/Components/WalletHeader.tsx
```diff
@@ -0,0 +1,80 @@
...
+  overview: {
+    available: boolean,
+    peerCount: string,
+    peerTitle: string,
+    threadCount: string,
+    threadTitle: string,
+    photoCount: string,
+    photoTitle: string
+  },
+  avatarUrl: number,
```
asutula: number?

andrewxhill: odd that got through

PR #427, opened 2018-08-29
Avatar

andrewxhill:
- Updates Avatar component to be more intelligent
- If given no peer_id, Avatar assumes you just want your own avatar
- Updates all uses of Avatar (except comments shit, @asutula will need to update that)
- Fixes the lost avatar creation during onboarding
- Fixes the lost default thread creation during onboarding
- Makes the app work again?


App/Containers/ThreadsList.js
```diff
@@ -64,17 +62,20 @@ class ThreadsList extends React.PureComponent {
   componentDidMount () {
     this.props.navigation.setParams({
       profile: this.props.profile,
+      online: this.props.online,
```
andrewxhill: Will always go for cafe while waiting

andrewxhill:
I take this part back, "If given no peer_id, Avatar assumes you just want your own avatar". I thought about it for a second and realized there will probably end up being some case where an avatar defaults to someone's own avatar because peer_id is undefined by mistake. seems like a bad confusion so added a flag that you must use if you just want your own avatar, 'owner'. See WalletHeader.tsx for example

App/Sagas/TextileSagas.ts
```diff
@@ -93,26 +94,43 @@ export function * updateNodeOverview ( action: ActionType<typeof TextileNodeActi
   }
 }
 
+export function * onboardedSuccess(action: ActionType<typeof PreferencesActions.onboardedSuccess>) {
```
asutula: Seems like we can just check for and possibly create the default thread after starting the node, before telling the rest of the app that the node is started. Also, that way we don't have to tell the app to refresh threads etc.

App/Sagas/TextileSagas.ts
```diff
@@ -93,26 +94,43 @@ export function * updateNodeOverview ( action: ActionType<typeof TextileNodeActi
...
   yield call(NavigationService.navigate, 'PrimaryNavigation')
-  yield take(getType(TextileNodeActions.startNodeSuccess))
 
-  let defaultThread: TT.Thread | undefined = yield call(getDefaultThread)
-  if (!defaultThread) {
-    yield put(ThreadsActions.addThreadRequest('default' as TT.ThreadName))
-    const action: ActionType<typeof ThreadsActions.addThreadSuccess> = yield take(getType(ThreadsActions.addThreadSuccess))
-    defaultThread = action.payload.thread
-    yield put(ThreadsActions.refreshThreadsRequest())
+  // PreferencesActions.onboardedSuccess will setup the node/default thread, so wait for those
```
asutula: Seems like something would probably be very broken if the node weren't online at this point. If that were the case, your loop that waits for it to come online will probably be looping forever. We should probably disable the UI for updating your avatar if the node isn't online.

App/Sagas/TextileSagas.ts
```diff
@@ -93,26 +94,43 @@ export function * updateNodeOverview ( action: ActionType<typeof TextileNodeActi
...
 
-  let defaultThread: TT.Thread = yield call(getDefaultThread)
+  // Ensure that our node is online before we try to update the profile
+  let online = yield select(TextileNodeSelectors.online)
+  while (!online) {
+    yield call(delay, 150)
+    online = yield select(TextileNodeSelectors.online)
+  }
 
+  let defaultThread: TT.Thread = yield call(getDefaultThread)
```
asutula: This logic is a little different than above, but same comment applies.

App/Sagas/TextileSagas.ts
andrewxhill: okay, added a ticket to move that https://github.com/textileio/textile-mobile/issues/428

App/Components/Avatar.tsx
```diff
@@ -0,0 +1,113 @@
...
+import {View, Image, ImageURISource, StyleProp, ViewStyle} from 'react-native'
+import TextileImage from '../../TextileImage'
+import { Profile } from '../Models/TextileTypes'
+
+import styles from './Styles/AvatarStyles'
+import {RootState} from '../Redux/Types'
+import {NavigationScreenProps} from 'react-navigation'
+
+
+interface Props extends NavigationScreenProps<{}> {
```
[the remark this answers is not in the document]
andrewxhill: ah, copy past, `Props extends React.PureComponent<{}>`?

App/Sagas/TextileSagas.ts
andrewxhill: okay, was mostly piling on the safety checks last night racking my brain why it was all broken. removing the online part

PR #438, opened 2018-08-29
Make sure default thread is created before startNodeSuccess

### LINKED ISSUE #428: Add default thread as part of first node startup cycle
remove `onboardedSuccess` saga and just add the default before letting the rest of the app node that the node started

asutula:
Ok i think this is simple and works. Idea is to make sure the default thread exists right after stating the node, but before dispatching the `startNodeSuccess` action. Since we don't really run any code until after the node is started, it is then safe to assume the default thread exists everywhere else. Fixes #428

### DIFF
--- a/App/Sagas/ThreadsSagas.ts
+++ b/App/Sagas/ThreadsSagas.ts
@@ -50,9 +50,9 @@ export function * acceptExternalInvite (action: ActionType<typeof ThreadsActions
   }
 }
 
-export async function getDefaultThread (): Promise<Thread | undefined> {
+export async function getDefaultThread (): Promise<Thread> {
   const threads = await TextileNode.threads()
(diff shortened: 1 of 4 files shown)


App/Sagas/ThreadsSagas.ts
```diff
@@ -50,9 +50,9 @@ export function * acceptExternalInvite (action: ActionType<typeof ThreadsActions
   }
 }
 
-export async function getDefaultThread (): Promise<Thread | undefined> {
+export async function getDefaultThread (): Promise<Thread> {
   const threads = await TextileNode.threads()
-  return threads.items.find(thread => thread.name === 'default')
+  return threads.items.find(thread => thread.name === 'default')!
```
asutula: We just assume the thread exists since we know we created it when the node was started, but before any sagas that use the default thread run

PR #445, opened 2018-08-30
Couple more more couple

andrewxhill:


App/Components/ProgressiveImage.tsx
```diff
@@ -12,7 +12,61 @@ export interface ProgressiveImageProps {
...
+      />
+      <TextileImage
+        imageId={this.props.imageId}
+        path={this.props.path}
+        style={[{backgroundColor: 'transparent', position: 'absolute', top: 0, bottom: 0, left: 0, right: 0}, this.props.style, this.state.androidPreview && {height: 0}]}
+        resizeMode={this.props.resizeMode}
+        capInsets={this.props.capInsets}
+        onLoad={this._androidLoad.bind(this)}
+      />
+    </Fragment>
```
andrewxhill: Android native textile image element wasn't setup to allow them to be nested like ios does. so went with the less time consuming fix here for now. works well enough.

App/Containers/AddCaptionScreen.js
```diff
@@ -67,9 +65,12 @@ class AddCaptionScreen extends React.Component {
   }
 
   _renderAsset () {
+    const sourceUri = this.props.image.origURL && this.props.image.origURL !== ''
+      ? this.props.image.origURL
+      : this.props.image.uri
     return (
       <Image
-        source={{uri: this.props.image.origURL}}
+        source={{uri: sourceUri, isStatic: true}}
```
andrewxhill: android likes this isStatic parameter or just sits there and doesn't do anything and wont tell you why

App/Containers/ThreadsList.js
```diff
@@ -46,7 +46,9 @@ class ThreadsList extends React.PureComponent {
       </TextileHeaderButtons>
     )
     const headerTitle = (
-      <Image style={navStyles.headerLogo} source={require('../SB/views/ThreadsList/statics/logo.png')} />
+      <View style={navStyles.headerCenter}>
```
andrewxhill: android also likes to left align all our headers. so a few fixes in here for those

App/Services/Notifications.ts
```diff
@@ -31,7 +31,7 @@ export function toPayload(notification: Notification): NotificationsPayload | un
   switch (notification.type) {
     case(NotificationType.receivedInviteNotification): {
       const title = 'New Invite'
-      const message =  [actor, notification.body, 'to', notification.subject].join(' ')
+      const message =  [actor, notification.body].join(' ')
```
andrewxhill: doh, gets rid of that double speak in one of the feed item types

andrewxhill:
Okay, this is gtg. Fixes a few issues around the fort

App/Containers/Wallet.js
```diff
@@ -19,16 +19,17 @@ class Wallet extends React.PureComponent {
     const params = navigation.state.params || {}
     const greeting = params.username ? 'Hello, ' + params.username : 'Hi there!'
     const headerTitle = (
-      <Text style={navStyles.headerTitle}>
-        {greeting}
-      </Text>
+      <View style={navStyles.headerCenter}>
```
asutula: There is a property on the navigationOptions called `headerTitleStyle` where you can pass in styles, including margin, padding, etc, and avoid having to do this view wrapper everywhere.

https://reactnavigation.org/docs/en/headers.html#adjusting-header-styles

App/Containers/ThreadsList.js
asutula: See my other comment about this.

App/Containers/Wallet.js
andrewxhill: i assumed something was getting lost here, because we are setting that property in the navigator creation step

asutula: I don't see that. I only see this:
```
headerTitle: {
    fontFamily: 'BentonSans-Bold',
    fontSize: 17,
    paddingTop: 8, // <- needed because BentonSans has that weird bottom gap
    color: Colors.charcoal
  }
```

asutula: Nothing about margin or horizontal anything

asutula: That is from `NavigationStyles.js`

andrewxhill: so it was never working? interesting

andrewxhill:
okay _now_ gtg

PR #447, opened 2018-08-30
no refresh spinner in Feed except as you are pulling down

andrewxhill:


andrewxhill:
the pull action still shows a spinner. no hack needed. bounces back up right after pull is over

asutula:
Yea but it only shows while you are pulling. If the refresh took 5 seconds to complete, the user would be left with impression that the refresh is done at that it finished immediately.

andrewxhill:
i'd lean toward just going with that

asutula:
I was just looking at the different places we have pull to refresh UI and what happens in each case. Might be worth reviewing together to make sure it's right. In many cases we're just calling refreshMessages which, based on our conversations yesterday, i'm still not sure is correct/needed.

asutula:
This is an easy fix to allow the user (and us, for debugging and understand how the app is performing) to actually know when a refresh is complete. Why wouldn't we want that?

andrewxhill:
i was under the impression the opposite, that that was the only thing ever needed in a - please force a refresh - scenario. 

longer term, i'm with you. shorter term, i've seen the dismay of releases where endless refresh spinners stick there and we aren't certain why. we'll need to spend time here investigating. 

on top of that, you'll need to update refreshmessages to return an empty 'done refreshing' event or i'm not sure how else you'll get it. 

or you could put a middle ground hack where it times out the ui refresh after so long if it doesn't get a message, but that's a hack so what's the point also.

final point in slack

PR #454, opened 2018-08-31
better external invites

andrewxhill:


andrewxhill:
okay, this one is gtg

andrewxhill:
okay, that should be gtg now

PR #455, opened 2018-08-31
Comments, Threads/Photos redux refactor

asutula:
Big idea is that the redux around viewing threads and photos should be managed by the same reducer. Makes the data flow much more simple.

Also consolidated redux list of threads and photo data by thread since they were redundant. 

Comments is working, needs some improvements for sure, but lets test it out.

Most functionality working, still need to test all far corners of the app, thread joining etc.


App/Redux/PhotoViewingRedux.ts
```diff
@@ -0,0 +1,197 @@
+import { createAction, ActionType, getType } from 'typesafe-actions'
```
asutula: This new redux should probably be renamed to "ThreadsRedux" if we totally dissolve the old `ThreadsRedux`. It combines a bunch of stuff from `ThreadsRedux`, `UIRedux` and `TextileReudx`. There was a lot of redundancy between those, so it ends up being less code like this.

This redux manages all things about loading, creating, and viewing threads and photos.

App/SB/views/ThreadDetail/index.js
```diff
@@ -109,9 +107,10 @@ class ThreadsDetail extends React.PureComponent {
     })
   }
 
-  _onPhotoSelect = () => {
-    return (photoId) => {
-      this.props.viewPhoto(photoId, this.props.threadId)
+  _onPhotoSelect = (photo: Photo) => {
+    return () => {
+      this.props.viewPhoto(photo.id)
+      this.props.navigation.navigate('PhotoViewer')
```
asutula: One simplifying change I made is to manage state about which thread/photo is being viewed or operated on in redux, and leave navigation up to the UI layer any time that is possible.

Previously, we were dispatching a redux action that would update some state in the reducer and we would also have a saga listening for the action and the saga would trigger a navigation. Super confusing and hard to follow. 

Now the UI dispatches a redux action to update state about which thread/photo we are viewing and then navigates to the appropriate screen. We don't have to pass any data along in the navigation because the screen we are navigating to can get it from redux.

asutula:
Lots of "changes" in this PR because I renamed the folder/file for the screen that displays the details of a thread from "Thread**s**Details" to "ThreadDetails". Such an tiny change making a mess here, but every time I worked near that code, the name would confuse me.

App/Containers/ThreadsList.js
```diff
@@ -119,7 +120,8 @@ class ThreadsList extends React.PureComponent {
 
   _onPressItem = (photo) => {
     const { id, name } = photo
-    this.props.viewThread(id, name)
+    this.props.viewThread(id)
```
asutula: This is setting the redux state of the currently viewed thread.

App/Containers/ThreadsList.js
```diff
@@ -119,7 +120,8 @@ class ThreadsList extends React.PureComponent {
...
+    this.props.navigation.navigate('ViewThread')
```
asutula: And this just simply navigates to the thread details screen. That screen gets it's data from redux only.

App/Models/TextileTypes.ts
```diff
@@ -126,7 +126,8 @@ export interface Photo {
   date: string
   author_id: PeerId
   caption?: string
-  metadata: PhotoMetadata
+  username?: string
+  metadata?: PhotoMetadata
```
asutula: Updated the to reflect the go layer model

App/Redux/PhotoViewingRedux.ts
```diff
@@ -0,0 +1,197 @@
...
+    readonly error?: string
+  }
+  readonly removingThread?: {
+    readonly id: ThreadId
+    readonly error?: string
+  }
+  readonly viewingWalletPhoto?: Photo,
+  readonly viewingThread?: Thread,
+  readonly viewingPhoto?: Photo,
+  readonly authoringComment?: string
```
asutula: The above 4 properties is where we store info about what thread and photo is being viewed or operated on. Should probably move sharing related state in here too, but haven't yet.

App/Redux/PhotoViewingRedux.ts
```diff
@@ -0,0 +1,197 @@
...
+  readonly photos: ReadonlyArray<Photo>
+  readonly error?: Error
+}
+
+type ThreadMap = {
+  readonly [key: string]: ThreadData
+}
+
+type PhotoViewingState = {
+  readonly threads: ThreadMap,
```
asutula: All data for listing threads and the photos within those threads is derived from this property.

App/Redux/PhotoViewingSelectors.ts
```diff
@@ -0,0 +1,25 @@
+import { RootState } from './Types'
+import { ThreadData } from './PhotoViewingRedux'
+
+export function defaultThreadData (state: RootState): ThreadData | undefined {
```
asutula: Kinda nice to move selectors into a module-wide file. They can get complicated. We should move a lot of the complicated code that exists in our `mapDispatchToProps` into selectors.

App/SB/components/CommentBox/CommentBox.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const { keyboardHeight, value, onUpdate, onSubmit } = props
+
+  return (
+    <View style={[styles.commentFooter, { bottom: keyboardHeight }]}>
+      <TextInput
+        style={styles.textInput}
+        placeholder='Write a comment...'
+        placeholderTextColor='#4a4a4a'
+        onChangeText={onUpdate}
+        value={value}
```
asutula: It's cool, somehow RN implements some components to be two-way-bound... `onChangeText` allows you to get updates, update your redux state or whatever, but `value` allows you to set the current text in the text input. We use it here to clear out the text input when our `authoringComment` state becomes `undefined`. Somehow they avoid a recursive death loop of events.

App/Components/ProgressiveImage.tsx
```diff
@@ -34,14 +34,14 @@ export default class ProgressiveImage extends React.Component<ProgressiveImagePr
       <TextileImage
         imageId={this.props.imageId}
         path={'thumb'}
```
andrewxhill: ah shoot, both of these 'thumbs' should just be `this.props.previewPath` if you want to pop that in

App/SB/components/CommentCard/index.tsx
```diff
@@ -1,29 +1,32 @@
...
+  // const uri = props.photo.author_id ? 'https://cafe.us-east-1.textile.io/ipns/' + props.photo.author_id + '/avatar' : undefined
+  // const username =  props.photo.username ? props.photo.username : props.photo.author_id.substring(0, 8).toUpperCase()
 
+  const { username, avatarUri, comment, date, subComments, isSubComment } = props
+  const dateString = moment.utc(date).fromNow()
+  const defaultSource = require('../../views/Settings/statics/main-image.png')
   return (
     <View style={[styles.comment, isSubComment ? styles.subComment : styles.withDivider ]}>
-      <Avatar style={{marginRight: 11}} width={38} height={38} uri={uri} defaultSource={defaultSource} />
+      <Avatar style={{marginRight: 11}} width={38} height={38} uri={avatarUri} defaultSource={defaultSource} />
```
andrewxhill: you'll just change this to `peer_id={peer_id}` now

App/SB/views/ThreadPhotoDetail/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+      isSubComment: false,
+      subComments: []
+    }
+  }
+  // TODO: comments should always be defined: https://github.com/textileio/textile-go/issues/270
+  const comments = viewingPhoto.comments || []
+  const commentCardProps = comments.slice().reverse().map(comment => {
+    const props: CommentCardProps = {
+      username: comment.username || 'unknown',
+      avatarUri: 'https://cafe.textile.io/ipns/' + comment.author_id + '/avatar',
```
andrewxhill: yeah, so just pass `peer_id: comment.author_id`

App/Sagas/ThreadsSagas.ts
```diff
@@ -57,55 +56,19 @@ export async function getDefaultThread (): Promise<Thread> {
 
 export function * pendingInvitesTask () {
   // Process any pending external invites created while user wasn't logged in
-  const threads = yield select(ThreadsSelectors.threads)
-  if (threads.pendingInviteLink) {
-    yield call(DeepLink.route, threads.pendingInviteLink, NavigationService)
+  const pendingInviteLink: string | undefined = yield select(ThreadsSelectors.pendingInviteLink)
+  if (pendingInviteLink) {
+    yield call(DeepLink.route, pendingInviteLink, NavigationService)
     yield put(ThreadsActions.removeExternalInviteLink())
```
andrewxhill: Redid this in my existing PR, but didn't want to touch ThreadsSagas until this beast was in master so just worked around it for now

andrewxhill: The pending invite probably _does_ want to hit the AuthRedux, since it can just store the referral code right in the form data that way. and i con't see pending thread invites being used outside the auth flow, so... let's revisit after this merges

App/Components/ProgressiveImage.tsx
asutula: Fixed

App/SB/views/ThreadPhotoDetail/index.tsx
asutula: Done

App/Services/EventHandlers/TextileNodeEventHandler.ts
```diff
@@ -19,14 +20,15 @@ export default class TextileNodeEventHandler {
     TextileNode.eventEmitter.addListener('onOnline', () => {
       this.store.dispatch(TextileNodeActions.nodeOnline())
     })
-    TextileNode.eventEmitter.addListener('onThreadUpdate', (payload) => {
-      this.store.dispatch(TextileNodeActions.getPhotoHashesRequest(payload.thread_id))
+    TextileNode.eventEmitter.addListener('onThreadUpdate', (update: Update) => {
+      const { type } = update.block
+      if (type === BlockType.CommentBlock || type === BlockType.LikeBlock || type === BlockType.PhotoBlock) {
```
asutula: We only refresh the thread for comments, likes, photos, and ignores

App/Services/EventHandlers/TextileNodeEventHandler.ts
```diff
@@ -19,14 +20,15 @@ export default class TextileNodeEventHandler {
...
+        this.store.dispatch(PhotoViewingActions.refreshThreadRequest(update.thread_id))
+      }
     })
-    TextileNode.eventEmitter.addListener('onThreadAdded', () => {
-      this.store.dispatch(ThreadsActions.refreshThreadsRequest())
+    TextileNode.eventEmitter.addListener('onThreadAdded', (payload) => {
     })
-    TextileNode.eventEmitter.addListener('onThreadRemoved', () => {
-      this.store.dispatch(ThreadsActions.refreshThreadsRequest())
+    TextileNode.eventEmitter.addListener('onThreadRemoved', (payload) => {
```
asutula: There is really nothing to do in responding to thread add/remove events. The app redux already knows about those events because the app is the thing that created/removed the thread. @sanderpick I guess this is the only thing I wanted to be sure is ok.

App/Services/EventHandlers/TextileNodeEventHandler.ts
```diff
@@ -19,14 +20,18 @@ export default class TextileNodeEventHandler {
...
       this.store.dispatch(TextileNodeActions.nodeOnline())
     })
-    TextileNode.eventEmitter.addListener('onThreadUpdate', (payload) => {
-      this.store.dispatch(TextileNodeActions.getPhotoHashesRequest(payload.thread_id))
+    TextileNode.eventEmitter.addListener('onThreadUpdate', (update: Update) => {
+      const { type } = update.block
+      if (type === BlockType.CommentBlock ||
+        type === BlockType.LikeBlock ||
+        type === BlockType.PhotoBlock ||
+        type === BlockType.IgnoreBlock) {
```
asutula: Only refresh for comments, likes, photos, and ignores.

asutula:
Ok, I think we're good to go here.

App/Models/TextileTypes.ts
```diff
@@ -219,3 +215,25 @@ export interface NotificationEngagement {
...
+  InviteBlock = 0,
+	ExternalInviteBlock = 1,
+	JoinBlock = 2,
+	LeaveBlock = 3,
+	PhotoBlock = 4,
+	CommentBlock = 5,
+	LikeBlock,
+
+	IgnoreBlock = 200,
+	MergeBlock  = 201
```
sanderpick: boom

App/Redux/PhotoViewingSelectors.ts
andrewxhill: nice, makes sense

android/app/build.gradle
```diff
@@ -7,6 +7,11 @@ buildscript {
     }
 }
 
+project.ext.envConfigFiles = [
+    debug: ".env.staging",
+    release: ".env.production"
+]
```
asutula: You have to tell the `react-native-config` module which config file to use for which build. If you don't, it defaults to `.env` which would explain that report we got.

PR #462, opened 2018-09-01
Quick take on blocking onvoarding check until redux is rehydrated

asutula:


App/Redux/StartupRedux.ts
```diff
@@ -8,4 +10,26 @@ const actions = {
 
 export type StartupAction = ActionType<typeof actions>
 
+export type StartupState = {
+  started: boolean
```
asutula: The startup action has always be fired when redux-persist rehydration is complete. Just adding some state to track when that happens so you can check for it in sagas.

App/Sagas/DeepLinkSagas.ts
```diff
@@ -29,6 +30,10 @@ export function * inviteAfterOnboard () {
 }
 
 export function * routeThreadInvite(url: string, hash: string ) {
+  const reduxStarted: boolean = yield select(startupSelectors.started)
+  if (!reduxStarted) {
+    yield take(getType(StartupActions.startup))
+  }
```
asutula: So basically, this function is currently being called before redux persist rehydration is complete. This logic blocks the saga until the startup/rehydration is complete.

PR #464, opened 2018-09-01
Cleans up all lint warnings

andrewxhill:
Okay @asutula only 221 files changed :D 

figured we could start adding a lint step on the build checks for a simple value/quality quick win here.


asutula:
Ok this is cool, just wondering the best way to integrate it... Tried a test merge on my comments branch, and that's just a mess. Wondering if we should get the comments branch merged and then, merge in the new tsconfig.json and then re-run the lint fix? Thoughts?

tslint.json
```diff
@@ -1,9 +1,52 @@
...
     "extends": [
-        "tslint:recommended"
+      "tslint:recommended",
+      "tslint-react"
     ],
+    "linterOptions": {
+      "exclude": [
+        "node_modules/**",
+        "typings/**",
+        "App/Sagas/**/*",
```
asutula: Not sure what order you created the two branches for this lint work, but this looks weird to me. My impression is that this branch should be linting all files. Not sure why these exclude items are here.

asutula: FYI... I did already merge the other lint branch into this one.

andrewxhill: I put those in there on purpose. It was so I didn't have to clean up all the tslint issues in those folders that you are changing so much in your branch. It was a best attempt to avoid merge hell. Then we can go back and selectively add them.

asutula: Ah ok cool. I thought that was the case in the other branch and this branch was to go all in. This works, tho

andrewxhill: yeah... would be a nightmare

PR #466, opened 2018-09-02
tslint

andrewxhill:


.circleci/config.yml
```diff
@@ -5,6 +5,12 @@ jobs:
         docker:
             - image: circleci/node:8
         steps:
+            - restore_cache: # If any cache exists, it'll way speedup these checkouts
+                keys:
+                    - source-v1-{{ .Branch }}-{{ .Revision }}
+                    - source-v1-{{ .Branch }}-
+                    - source-v1-
+
```
andrewxhill: Found that moving the project_workspace (creating mostly) was way more time consuming than just caching our git and restoring + checkout. also with the above setup, it'll work for a brand new branch, skipping the really time consuming 5gb clone

.circleci/config.yml
```diff
@@ -15,37 +21,45 @@ jobs:
...
+                    - node_modules
 
-            - persist_to_workspace:
-                root: ~/textile-mobile
+            # create a cache entry which will avoid checking out all 5GB of subsequent builds
+            - save_cache:
+                key: source-v1-{{ .Branch }}-{{ .Revision }}
                 paths:
-                    - node_modules
+                    - ".git"
```
andrewxhill: here's where our repo is cached for all future builds. (and later stages of this build)

lint / tests go right of the old cache so they can just run in parallel

tslint.json
```diff
@@ -1,9 +1,42 @@
...
+    "linterOptions": {
+      "exclude": [
+        "node_modules/**",
+        "typings/**",
+        "App/Sagas/**/*",
+        "App/Redux/**/*",
+        "App/Navigation/**/*",
+        "App/Models/**/*",
+        "App/SB/**/*",
+        "**/*.json"
```
andrewxhill: Okay, kind of a weak first go here. But i wanted to avoid having to basically PR a file in pretty much every folder of the repo on the first go. It also should save us some headaches merging with the Comments PR still outstanding. We can add these in later.

tslint.json
```diff
@@ -1,9 +1,42 @@
...
+      "variable-name": [true, "ban-keywords", "check-format", "allow-pascal-case", "allow-leading-underscore"],
+      "trailing-comma": [
+        true,
+        {
+          "multiline": "never",
+          "singleline": "never"
+        }
+      ],
+      "semicolon": [true, "never"],
+      "max-line-length": [true, 200]
```
andrewxhill: A lot of these rules probably aren't the greatest. But again, I wanted to get v0 of this PR in without having to change the entire project off the bat.

andrewxhill:
finnnnally new circleci stuff works. split out to work a bit faster and include both lint and tslint

andrewxhill:
Ah, this merges into https://github.com/textileio/textile-mobile/pull/464 :) 

needed one before the other

PR #469, opened 2018-09-05
Investigate node cycle

### LINKED ISSUE #467: Node keeps getting stuck as stopped
Only way to fix it is hard-kill. Happens about once every 2-3 days to me, so seems critical to try and fix.

In the settings screen,

file 2

and verbose

file 1

asutula:
- Possibly fixed a bug that would cause the node to sometimes not start.
- Display notifications about what's going on with the node in the background if `verboseUi` is `true`
- Display app/node state message in `RootContainer` so it's visible no matter which screen you're on

Fixes #467 I hope

### DIFF
--- a/App/Sagas/NodeLifecycle.ts
+++ b/App/Sagas/NodeLifecycle.ts
@@ -4,14 +4,17 @@ import { ActionType, getType } from 'typesafe-actions'
 import RNFS from 'react-native-fs'
 import Config from 'react-native-config'
 import BackgroundTimer from 'react-native-background-timer'
+import PushNotification from 'react-native-push-notification'
 
-import TextileNodeActions, { TextileNodeSelectors, NodeState } from '../Redux/TextileNodeRedux'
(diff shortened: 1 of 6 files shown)


App/Sagas/NodeLifecycle.ts
```diff
@@ -4,25 +4,35 @@ import { ActionType, getType } from 'typesafe-actions'
...
-      // Get our current node state and create/start the node if it doesn't exist or is stopped
+      if (verboseUi) {
+        yield call(displayNotification, 'App State Change: ' + action.payload.newState)
+      }
+
+      // Get our current node state and create/start the node if it isn't started
       // Use fork so we don't block listening for the next app state change while the node is created and started
       const nodeState: NodeState = yield select(TextileNodeSelectors.nodeState)
-      if (nodeState === NodeState.nonexistent || nodeState === NodeState.stopped) {
+      if (nodeState !== NodeState.started) {
```
asutula: This is the possible bug. Node wouldn't be started if it were in some transitioning state like `stopping`.

PR #472, opened 2018-09-05
try @textile/go-mobile for frameworks

sanderpick:
Couldn't resist trying this... seems to work just fine.


asutula:
How far should we take this for now? Just this will make updating the frameworks much easier (no more copy/paste from textile-go to textile-mobile). We could also make the publish the JS/TS layer as well?

sanderpick:
Sweet. We can also purge all the old framework versions from the repo... should make the git repo caching faster?

sanderpick:
Def can do all the TS types and the actual TextileNode.ts etc. there too... figured this was a good place to start. I'll push the repo up and can get it publishing from the textile-go CI at some point soon

PR #473, opened 2018-09-05
All tslint

asutula:
That was fun


andrewxhill:
you are in luck, looks like the fun continues

asutula:
Ha ok, should be fixed now, let's see.

asutula:
There we go @andrewxhill all fixed

PR #476, opened 2018-09-06
Improved comments UX

asutula:
- Created a general use keyboard aware component that resizes with keyboard show/hide in an animated fashion
- Using that component for comments screen
- Comment input box now animates up and down smoothly and is aligned with the top of the keyboard
- Scroll to bottom of screen when newly added comment is visible


App/Components/KeyboardResponsiveContainer.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+interface Props {
+  style?: ViewStyle
+}
+
+interface State {
+  height: number
+}
+
+export default class KeyboardResponsiveContainer extends React.Component<Props, State> {
```
asutula: We can use this anywhere we want the container view to resize with the keyboard.

App/SB/views/ThreadPhotoDetail/index.tsx
```diff
@@ -54,13 +55,28 @@ class ThreadPhotoDetail extends Component<Props, State> {
...
   }
 
+  scrollToEnd = () => {
+    if (this.scrollView) {
+      this.scrollView.scrollToEnd()
+    }
+  }
+
+  componentDidUpdate (previousProps: Props, previousState: State) {
+    if (this.props.commentCardProps.length > previousProps.commentCardProps.length) {
```
asutula: Scroll to the bottom of the comments as soon as the new comment is there to see

App/Components/KeyboardResponsiveContainer.tsx
sanderpick: very cool

PR #479, opened 2018-09-08
Storage options + pin all photos

andrewxhill:
Still working on this, but wanted to start sharing the flow that seems to work


App/Redux/PreferencesRedux.ts
```diff
@@ -125,6 +151,12 @@ export function reducer (state: PreferencesState = initialState, action: Prefere
       service.status = action.payload.status === undefined ? !service.status : action.payload.status
       return {...state, services: {...state.services, [action.payload.name]: service}}
     }
+    case getType(actions.toggleStorageRequest): {
+      const storageType = state.storage[action.payload.name]
+      if (!storageType) { return state }
+      storageType.status = action.payload.status === undefined ? !storageType.status : action.payload.status
+      return {...state, storage: {...state.storage, [action.payload.name]: storageType}}
```
andrewxhill: Todo: we can set the first local photo date the moment that the option is toggled. in that way we'll never get photos older than the option change. you can also toggle it off and then back on again and we'll skip all that happened in between

App/Redux/ProcessingImagesRedux.ts
```diff
@@ -80,14 +86,17 @@ export interface ProcessingImage {
 }
 
 export interface ProcessingImagesState {
-  readonly images: ProcessingImage[]
+  readonly images: ProcessingImage[],
+  readonly lastPhotoRefresh: number // epoch in seconds
```
andrewxhill: This is that date we need to set. We can update it when the option is toggled and whenever we run the query.

andrewxhill: oop, also changed this to milliseconds. just kept everything in JS as milliseconds until the moment it makes the objc request in seconds

ios/NativeModules/TextileNode.m
```diff
@@ -47,6 +48,77 @@ - (dispatch_queue_t)methodQueue {
...
+
+      // Limit to only photo media types
+      // NSPredicate *mediaPhotos = [NSPredicate predicateWithFormat:@"mediaType = %i", PHAssetMediaTypeImage];
+      // Limit to only photos modified later than or equal to our specified date
+      NSPredicate *minDate = [NSPredicate predicateWithFormat:@"modificationDate > %@", epochNSDate];
+      // Limit to camera roll?
+      // NSPredicate *cameraRoll = [NSPredicate predicateWithFormat:@"modificationDate > %@", epochNSDate];
+      onlyImagesOptions.predicate = minDate;
+      // Combine the multiple predicates (didn't seem to work combining them like this)
+      // [NSCompoundPredicate orPredicateWithSubpredicates:@[mediaPhotos, minDate]];
```
andrewxhill: Need to return to these after testing live. Not totally clear what kind of duplication happens over galleries

App/Sagas/StorageSagas.ts
```diff
@@ -0,0 +1,63 @@
...
+    // Issue: if the user doesn't want to store private files on remote IPFS, we need to record that these are
+    // only available locally in case the user then shares
+    // Idea perhaps...
+    // Store a table of images in StorageRedux
+    // path string, hash string, remote string
+    // you could set that in this sage with blockId and photo.path above
+    // whenever a user shares a photo to a thread from the wallet, just grab the remote?
+    // if not remote, do a photo remote pin
+    // it will take a slightly modified photo remote pin, since at the end of it all you will skip adding to the wallet
+    // maybe?
```
andrewxhill: Okay, I'd say the above is the biggest remaining question before testing this out. (also the go thing about orientations, but wouldn't be a breaker for testing with camera images)

ios/NativeModules/TextileNode.m
andrewxhill: moving a single file around galleries doesn't seem to create duplices (or change modified dates) so that's solid

andrewxhill:
@asutula would you be willing to jump into this branch and see about integrating it into your thread share flow? the trick is simply that 

1. wallet photos will be pinned to wallet before any remote pinning, so need to do a remote pin from an existing wallet which is different than the current setup
2. for remote pin, we may want to do it right from the wallet data and not the path data, since path could go away

andrewxhill:
this rough stab worked, https://github.com/textileio/textile-mobile/pull/481 but just isn't right i don't think

asutula:
Yea, can do.

ios/NativeModules/TextileNode.m
```diff
@@ -47,6 +48,73 @@ - (dispatch_queue_t)methodQueue {
...
+              // Send our event back to RN
+              [Events emitEventWithName:@"newLocalPhoto" andPayload:jsonStr];
+            }
+          }
+        }];
+
+      }];
+    }
+  }
+  resolve(@YES);
```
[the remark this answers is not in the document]
asutula: Cool, that is surprisingly clean. Just wondering if there's a reason you're not resolving the promise with the array of new photo data... Seems more simple to just return the data from the promise instead of emitting an event for each new photo, having to subscribe to that event in RN, etc.

asutula:
We can talk about this on the call, but just want to be sure I understand the use case around adding photos to the wallet without remote pinning. As a user, why would I want to do that?

App/Redux/PreferencesRedux.ts
andrewxhill: ah, this is in there now actually. no longer todo

ios/NativeModules/TextileNode.m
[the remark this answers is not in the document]
andrewxhill: wasn't sure how to wait for the N completion handlers and recompile their results into a single array of photos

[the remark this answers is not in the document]
asutula: Cool, there are ways. I can do that.

ios/NativeModules/TextileNode.m
```diff
@@ -47,6 +48,77 @@ - (dispatch_queue_t)methodQueue {
...
+
+
+      PHFetchResult *allPhotosResult = [PHAsset fetchAssetsWithMediaType:PHAssetMediaTypeImage options:onlyImagesOptions];
+      [allPhotosResult enumerateObjectsUsingBlock:^(PHAsset *asset, NSUInteger idx, BOOL *stop) {
+
+        // Get our path right away while we are in Obj-c
+        [asset requestContentEditingInputWithOptions:nil completionHandler:^(PHContentEditingInput * _Nullable contentEditingInput, NSDictionary * _Nonnull info) {
+
+          // get rid of file://
+          NSString *path = [contentEditingInput.fullSizeImageURL.absoluteString substringFromIndex:7];
```
asutula: Is this now a path that we "own" and are responsible for cleaning up? I think so, just want to get your take on it @andrewxhill

andrewxhill: I was under the impression that it was not something we needed to clean...

asutula:
Moved to PR 480

PR #489, opened 2018-09-11
Pin all photos from camera roll

asutula:
Had to move to this branch because of large file issues with @andrewxhill's original


asutula:
Ok this branch now puts all new camera roll photos into the wallet. You can test it out as is.

Next step is to create some UI that shows the in-progress image adds in the wallet.

asutula:
Oh, only tested for iOS so far. Need to see how it works on Android. It probably doesn't work ha.

App/Redux/ProcessingImagesRedux.ts
```diff
@@ -1,11 +1,10 @@
 import { createAction, ActionType, getType } from 'typesafe-actions'
-import { ThreadId, SharedImage, AddResult, BlockId } from '../Models/TextileTypes'
+import {ThreadId, SharedImage, AddResult, BlockId, ILocalPhotoResult} from '../Models/TextileTypes'
 import { RootState } from './Types'
 
 const actions = {
   insertImage: createAction('processingImages/INSERT_IMAGE', (resolve) => {
-    return (uuid: string, sharedImage: SharedImage, destinationThreadId: ThreadId, comment?: string) => resolve({ uuid, sharedImage, destinationThreadId, comment })
-    return (uuid: string, sharedImage: SharedImage, destinationThreadId: ThreadId, comment?: string) => resolve({ uuid, sharedImage, destinationThreadId, comment })
+    return (uuid: string, sharedImage: SharedImage, destinationThreadId?: ThreadId, comment?: string) => resolve({ uuid, sharedImage, destinationThreadId, comment })
```
asutula: Made `destinationThreadId` optional. In this case, the sagas for adding/sharing images will complete once the photo is added to the wallet. This is the case for photos added automatically from the camera roll.

asutula:
Ok I think this branch is safe to merge in so we can test auto pin internally. At this point, there is no UI in the wallet view to represent the photo after it's been queried from your camera roll, before it's been added to your wallet (no UI while uploading). So the photo will just appear once it's remote pinned and added to your wallet. 

This also isn't currently working on Android.

PR #495, opened 2018-09-12
bump go-mobile@0.1.9

sanderpick:
Yet another framework update.

- Adds a panic catch (that may or may not work, need to see) for this crash: https://fabric.io/we-are-set-inc/ios/apps/io.textile.wallet/issues/5b982b6d6007d59fcd61a83a?time=last-thirty-days
- Does not set services like `TextileService` to `nil` on `Stop()`... have a hunch this may be the root cause of a few of these crashes from trying to access de-allocated objects (`EXC_BAD_ACCESS KERN_INVALID_ADDRESS`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "tslint": "tslint -c tslint.json --project ."
   },
   "dependencies": {
-    "@textile/go-mobile": "^0.1.8",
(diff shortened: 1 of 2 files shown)


sanderpick:
Wanna try and get this into the store release, so gonna merge -- not a lot to review here anyway.

PR #502, opened 2018-09-12
Only run one platform build on master

andrewxhill:
We have an ios and ios-beta task (ditto android) that are both run on master. This should just move the ios task to only non-master and the ios-beta task will include the fastlane test and does the whole ios platform task in one go instead of 2


asutula:
Think it's a good idea to run tests on the results of a file merge in master? Seems like a good idea to me.

andrewxhill:
All the same tests are run for all branches. 

* Just the ios (general) and android (general) tasks are no longer run on Master. 
* The ios-beta and android-beta tasks are still only run on Beta but also include the fastlane tests that are part of the general tasks that are no-longer run. It allows us not to do the platform builds 4 times on each master CI pipeline, instead just doing 2 (one for each platform)

asutula:
I think I understand that idea, but I don't think it avoids having to build the code twice for each platform. For example, in iOS, tests are run in the simulator, so that requires a compile specific to the simulator. Then you compile for distribution to the app store which requires building for a real devices architecture.

asutula:
Our native tests at this point are almost non-existent, so the best optimization would be to just disable native tests for all branches.

andrewxhill:
Ah, gotcha, this may not be the right way then, take a peek. I thought we could do both with the same build.

andrewxhill:
looks gtg now

.circleci/config.yml
```diff
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
```
asutula: This stays `2`. It's not our version, it's the version of circle ci.

andrewxhill: yeah, "Enable build processing (preview)

Enable support for new config 2.1 features along with API build triggering and auto-cancellation of redundant workflows."

asutula: Ok, sounds dangerous. Let's see what happens. Mostly just need to watch deploys to App Store and Google Play. Not sure what happens if you cancel in the middle of submitting and how that will effect our build pipeline.

andrewxhill: Doesn't happen on Master, all commits are built on master

PR #503, opened 2018-09-12
require lint/tests before andriod/ios deploy. try to fix badge

andrewxhill:
We _do_ still have lint/tests in master build. So make them a pre-req for the deploy still. Doh

Hopefully fixes the badge on readme


asutula:
@andrewxhill just stuck the version number bumps in here

PR #518, opened 2018-09-13
Captions fix

asutula:
Fixes bug where caption text was a narrow view and threw off layout.

Also updates fastlane version


App/SB/components/CommentCard/statics/styles.js
```diff
@@ -12,7 +12,8 @@ export default StyleSheet.create({
   },
   content: {
     flexDirection: 'row',
-    margin: 12
+    width: '100%',
+    padding: 12
```
asutula: Using `padding` here because using `margin` was behaving strangely. It would put the correct margin on the top, left and bottom, but no margin on the right side. Ever seen that before @andrewxhill ?

PR #520, opened 2018-09-14
Comments in thread

### LINKED ISSUE #497: 'View all N comments' in image card
Multiple users have not noticed that you can comment on photos, because this UI is 'hidden' until you actually tap on a photo. Perhaps adding a comment count, or some indicator that there are comments, would be useful.

### LINKED ISSUE #509: Display recent comments in image card
Imagining some logic like "display the most recent two comments, limit each comment to two lines"

Only display username next to comments, no avatar.

asutula:
Closes #497 
Closes #509 

Displays up to two recent comments in the thread image, limited to one line each. Also displays "See all N comments...".

### DIFF
--- a/App/Models/TextileTypes.ts
+++ b/App/Models/TextileTypes.ts
@@ -125,7 +125,7 @@ export interface Photo {
   username?: string
   metadata?: PhotoMetadata
   comments: UserComment[]
-  Likes: UserLike[]
+  likes: UserLike[]
 }
 
 export interface ExternalInvite {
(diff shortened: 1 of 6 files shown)


App/Models/TextileTypes.ts
```diff
@@ -125,7 +125,7 @@ export interface Photo {
   username?: string
   metadata?: PhotoMetadata
   comments: UserComment[]
-  Likes: UserLike[]
+  likes: UserLike[]
```
asutula: No idea how this was working, must have been in JS code where the TS compiler wasn't checking. Or on @andrewxhill's computer where TS doesn't even check.

Issue #513, opened 2018-09-12
Placeholder text for feed view when empty


asutula:
(it's just blank white right now)

Should this be something like we display for the wallet and threads tours?

asutula:
Could be a juicy one for you @carsonfarmer

carsonfarmer:
We do currently have this similar to wallet and threads tours. I think the issue here is that, once you tap 'cool', there's nothing there... but this is what we do with the thread view as well. It is blank until a thread has been created or joined. Maybe we should add some text for both of these views?

asutula:
Just figured if there are no photos in a thread, we should display this placeholder content and that content should have a button that says "share photo" or point the user to the share photo button in the nav bar.

asutula:
So this isn't some tour screen they can dismiss. It will always display if there are no photos in the thread.

carsonfarmer:
From discussion on slack:
> actually, i guess its multiple things
> we should have placeholder text with 'action' buttons for all of them
> just went through:

-[] empty thread list is blank
-[] empty thread is blank
-[] empty feed view is blank

There are already SB designs for these, so we can use those.

carsonfarmer:
Update: the above views are all setup, but leads to quick flash when firing up because we don’t persist thread/photo data, and the queries to grab thread data are async. So checking if a thread list is empty will show up as 0 for a split second while that query happens. 
'Solution' seems to be to make the `treads` redux property optional, so then we'd interpret `undefined` as “doesn’t yet exist”... working on that now

carsonfarmer:
Fixed with #521

PR #521, opened 2018-09-14
Add placeholders for empty thread and notifications lists

carsonfarmer:
Nothing too fancy here, just adds placeholder text and icons (whose content are up for discussion).


App/Containers/ThreadsList.js
```diff
@@ -17,6 +18,13 @@ import PreferencesActions from '../Redux/PreferencesRedux'
 import TextileNodeActions from '../Redux/TextileNodeRedux'
 import { getThreads } from '../Redux/PhotoViewingSelectors'
 
+function isEmpty(obj) {
```
carsonfarmer: This is used to check if we have an empty object (the default). Couldn't find this one already built in... apparently this is safe than `Object.keys(obj).length`?

App/Containers/ThreadsList.js
```diff
@@ -86,14 +94,21 @@ class ThreadsList extends React.PureComponent {
       })
     }
     if (
+      this.props.threads &&
```
carsonfarmer: Check that we have an object, a previous object, and that they aren't the same

App/Containers/ThreadsList.js
```diff
@@ -86,14 +94,21 @@ class ThreadsList extends React.PureComponent {
...
       this._notificationPrompt()
     }
   }
 
   _notificationPrompt () {
-    if (this.props.notificationsPrompt && !this.props.tourScreen && this.props.threads.length > 0) {
+    if (
+      this.props.notificationsPrompt &&
+      !this.props.tourScreen &&
+      this.props.threads &&
```
carsonfarmer: Again, check for empty object and empty array.

App/Containers/ThreadsList.js
```diff
@@ -138,6 +153,50 @@ class ThreadsList extends React.PureComponent {
     )
   }
 
+  _renderList () {
```
carsonfarmer: Pulls this out for a cleaner conditional rendering step.

App/Containers/ThreadsList.js
```diff
@@ -138,6 +153,50 @@ class ThreadsList extends React.PureComponent {
...
+            onRefresh={this._onRefresh}
+            initialNumToRender={4}
+          />
+        </View>
+      )
+    }
+    return null
+  }
+
+  _renderPlaceholder () {
```
carsonfarmer: Only render placeholder if we either have an empty threads object, or the threads array is empty.

App/Containers/ThreadsList.js
```diff
@@ -190,30 +238,33 @@ class ThreadsList extends React.PureComponent {
...
-        thread.userCount = thread.photos.length > 0 ? [...new Set(thread.photos.map(photo => photo.author_id))].length : 1
-        // latest update based on the latest item
-        thread.updated = thread.photos.length > 0 && thread.photos[0].date ? Date.parse(thread.photos[0].date) : 0
-        // latest peer to push to the thread
-        thread.latestPeerId = thread.photos.length > 0 && thread.photos[0].author_id ? thread.photos[0].author_id : undefined
-      }
-      return thread
-    })
-    .sort((a, b) => a.updated < b.updated)
+  let threads = undefined
```
carsonfarmer: We'll now return undefined if our threads object from getThreads is empty

App/SB/views/FeedList/index.js
```diff
@@ -41,7 +42,7 @@ class Notifications extends React.PureComponent {
         />
       </HeaderButtons>
     )
-    const headerTitle = 'Feed'
+    const headerTitle = 'Notifications'
```
carsonfarmer: Sander and I talked about this briefly on slack... adding here, feel free to reject this one :)

App/SB/views/FeedList/index.js
```diff
@@ -110,6 +111,23 @@ class Notifications extends React.PureComponent {
     )
   }
 
+  _renderPlaceholder() {
+    if (this.props.notifications.length === 0 && this.props.showTourScreen !== true) {
```
carsonfarmer: Since loading from fresh shouldn't ever start with this screen (I think), no need to check for empty object like in thread view? Am I right in assuming this?

App/SB/views/ThreadDetail/index.js
```diff
@@ -128,6 +129,22 @@ class ThreadDetail extends React.PureComponent {
 
   _keyExtractor = (item, index) => item.id + '_' + index
 
+  _renderItems = () => {
```
carsonfarmer: Just pulls this out for cleaner implementation below.

carsonfarmer:
Fixes #513.

carsonfarmer:
@asutula, I didn't end up needing to allow threads to be undefined, this seemed a lot easier to deal with. But let me know if the undefined path seems like a better way to go?

App/SB/views/FeedList/index.js
asutula: I like it

App/Containers/ThreadsList.js
```diff
@@ -17,6 +18,13 @@ import PreferencesActions from '../Redux/PreferencesRedux'
 import TextileNodeActions from '../Redux/TextileNodeRedux'
 import { getThreads } from '../Redux/PhotoViewingSelectors'
 
+function isEmpty (obj) {
```
asutula: Don't think this is needed. See comment below.

App/Containers/ThreadsList.js
asutula: Don't need to check that this exists. See comment below.

App/Containers/ThreadsList.js
```diff
@@ -190,30 +238,33 @@ class ThreadsList extends React.PureComponent {
...
-        thread.userCount = thread.photos.length > 0 ? [...new Set(thread.photos.map(photo => photo.author_id))].length : 1
-        // latest update based on the latest item
-        thread.updated = thread.photos.length > 0 && thread.photos[0].date ? Date.parse(thread.photos[0].date) : 0
-        // latest peer to push to the thread
-        thread.latestPeerId = thread.photos.length > 0 && thread.photos[0].author_id ? thread.photos[0].author_id : undefined
-      }
-      return thread
-    })
-    .sort((a, b) => a.updated < b.updated)
+  let threads
```
asutula: Just initialize this to `[]` and then you don't have to check that `this.props.threads` is defined in many places in this file.

App/Containers/ThreadsList.js
```diff
@@ -190,30 +238,33 @@ class ThreadsList extends React.PureComponent {
...
+  if (!isEmpty(allThreads)) {
```
asutula: `allThreads` is just a `Thread[]`. No need for the `isEmpty` function. Just use `allThreads.length > 0`

App/Redux/PhotoViewingSelectors.ts
```diff
@@ -1,13 +1,14 @@
...
+import { Thread } from '../Models/TextileTypes'
 
 export function defaultThreadData (state: RootState): ThreadData | undefined {
   return Object.keys(state.photoViewing.threads)
     .map((key) => state.photoViewing.threads[key] )
     .find((threadData) => threadData.thread.name === 'default')
 }
 
-export function threadDataByThreadId (state: RootState, id: string) {
+export function threadDataByThreadId (state: RootState, id: string): ThreadData {
```
asutula: You can leave these return types I guess, but they are not needed since they are inferred by the TS compiler.

App/Containers/ThreadsList.js
asutula: For this case as well as `renderPlaceholder` and `renderList`, it might be a lot cleaner to just have props on this class like `showNotification: boolean`, `showPlaceholder: boolean` (you probably don't need `showList` because it's the opposite of `showPlaceholder`?). Anyway, all the logic to calculate those would be in `mapStateToProps` and not in the UI. Then the `render` function could just use simple logic, checking those props, to decide what to render.

App/Redux/PhotoViewingSelectors.ts
carsonfarmer: They help me reason about what's being returned, so I left them in... make it more readable to me?

asutula: Totally, it's fine to leave.

App/Containers/ThreadsList.js
carsonfarmer: Yup good call

App/Containers/ThreadsList.js
carsonfarmer: That actually doesn't work, and is what leads to the flashes. Because then there's no way to distingish between an empty array, and one that just hasn't finished populating yet. But if we leave this as `undefined`, then check if it has been replaced by an empty array, we know for sure its finished and just empty.

App/Containers/ThreadsList.js
carsonfarmer: Yes good idea, I've now done that for both views, much cleaner. Takes the logic out of the render functions and moves it to `mapStateToProps` as you suggested.

App/Containers/ThreadsList.js
carsonfarmer: See my response to that comment. We'll still need to check this, though indirectly now.

App/Containers/ThreadsList.js
carsonfarmer: Yup you're right, removed.

App/Containers/ThreadsList.js
```diff
@@ -138,50 +140,66 @@ class ThreadsList extends React.PureComponent {
...
+        <Text style={styles.emptyStateText}>
+          Nothing to see here yet... Start sharing your memories with friends and family with threads.
+            Create one using the <Icons name='add-thread' size={16} color='black' /> up top!
+          </Text>
+      </View>
+    )
+  }
+
   _renderTour () {
-    if (this.props.tourScreen === true) {
```
carsonfarmer: pulls conditional statements out...

App/Containers/ThreadsList.js
```diff
@@ -138,50 +140,66 @@ class ThreadsList extends React.PureComponent {
...
-              data={this.props.threads}
-              keyExtractor={this._keyExtractor}
-              renderItem={this._renderItem}
-              refreshing={this.props.refreshing}
-              onRefresh={this._onRefresh}
-              initialNumToRender={4}
-            />
-          </View>
-        )}
+        {this.props.showTourScreen && this._renderTour()}
```
carsonfarmer: ...and moves them down here

App/Containers/ThreadsList.js
```diff
@@ -190,38 +208,42 @@ class ThreadsList extends React.PureComponent {
...
+      })
+      .sort((a, b) => a.updated < b.updated)
+  }
 
   return {
     profile,
     threads,
     refreshing: !!state.ui.refreshingMessages,
-    notificationsPrompt: state.preferences.tourScreens.notifications,
+    showNotificationsPrompt: state.preferences.tourScreens.notifications && threads,
```
carsonfarmer: just changing for consistency with other booleans

App/Containers/ThreadsList.js
```diff
@@ -190,38 +208,42 @@ class ThreadsList extends React.PureComponent {
...
     services: state.preferences.services,
-    tourScreen: state.preferences.tourScreens.threads // <- default off so users don't see a screen flash,
+    showTourScreen: state.preferences.tourScreens.threads && threads, // <- default off so users don't see a screen flash,
```
carsonfarmer: moved boolean logic out of render functions to here

App/SB/views/FeedList/index.js
```diff
@@ -168,7 +190,7 @@ const mapDispatchToProps = (dispatch) => {
     refreshNotifications: () => dispatch(NotificationsActions.refreshNotificationsRequest()),
     readAllNotifications: () => dispatch(NotificationsActions.readAllNotificationsRequest()),
     refreshMessages: () => { dispatch(TextileNodeActions.refreshMessagesRequest()) },
-    clickNotification: (notification: Notification) => dispatch(NotificationsActions.notificationSuccess(notification)),
```
carsonfarmer: my linter was complaining about types in js files...

carsonfarmer:
Good for another round @asutula

carsonfarmer:
Ok, this has been synched with master... should still be gtg. Needs one more review (from @asutula?) to confirm my responses to his initial review.

App/Containers/ThreadsList.js
asutula: Makes sense

carsonfarmer:
No problemo @asutula. Going to update and merge.

PR #528, opened 2018-09-15
Sync photo roll

andrewxhill:


App/Containers/Wallet.js
```diff
@@ -171,7 +170,7 @@ const mapStateToProps = (state: RootState) => {
 const mapDispatchToProps = (dispatch) => {
   return {
     viewWalletPhoto: (photoId) => { dispatch(PhotoViewingActions.viewWalletPhoto(photoId)) },
-    refresh: (threadId: string) => { dispatch(PhotoViewingActions.refreshThreadRequest(threadId)) },
+    refresh: () => { dispatch(StorageActions.refreshLocalImagesRequest()) },
```
andrewxhill: It's the only refresh action that really makes sense in the wallet...

App/Models/TextileTypes.ts
```diff
@@ -220,7 +220,8 @@ export interface ILocalPhotoResult {
   modificationDate: string,
   orientation: number,
   path: string,
-  uri: string
+  uri: string,
+  canDelete: boolean
```
andrewxhill: This is now part of the payload that comes from the native layer. This way, ios/android can handle it differently as is probably the case, and if ever in the future one changes it doesn't need to be hunted down.

ios/NativeModules/TextileNode.m
```diff
@@ -85,25 +85,31 @@ - (dispatch_queue_t)methodQueue {
...
+            // Get the original filename to keep it straight on the system.
+            // Alternatively could probably just use timestamp, but this doesn't seem to have any lag
+            NSArray *resources = [PHAssetResource assetResourcesForAsset:asset];
+            NSString *orgFilename = ((PHAssetResource*)resources[0]).originalFilename;
+
+            // Get our path in the tmp directory
+            NSString *path = [[NSTemporaryDirectory()stringByStandardizingPath] stringByAppendingPathComponent:orgFilename];
+
+            // Write the data to the temp file
+            [imageData writeToFile:path atomically:YES];
```
andrewxhill: here it is, make a copy in our tmp/dir

andrewxhill:
okay, this works on device. love seeing my photos pop in there

andrewxhill:
should be gtg

App/Models/TextileTypes.ts
carsonfarmer: Yeh nice, good call

ios/NativeModules/TextileNode.m
```diff
@@ -85,25 +85,31 @@ - (dispatch_queue_t)methodQueue {
...
             NSDate *creationDate = asset.creationDate;
             // dataWithJSONObject cannot include NSDate
             NSString *dateString = [dateFormatter stringFromDate:newDate];
             NSString *creationDateString = [dateFormatter stringFromDate:creationDate];
+            // get an int
+            NSNumber *orientation = imageOrientation ? [NSNumber numberWithInt:imageOrientation] : [NSNumber numberWithInt:1];
 
-            // Create an event paylod
-            NSDictionary *payload = @{ @"uri": uri, @"path": path, @"modificationDate": dateString, @"creationDate": creationDateString, @"assetId": asset.localIdentifier, @"orientation": orientation};
+            NSDictionary *payload = @{ @"uri": path, @"path": path, @"modificationDate": dateString, @"creationDate": creationDateString, @"assetId": asset.localIdentifier, @"orientation": orientation, @"canDelete": @true};
```
asutula: Any java update for this needed?

ios/NativeModules/TextileNode.m
```diff
@@ -85,25 +85,31 @@ - (dispatch_queue_t)methodQueue {
...
+        [[PHImageManager defaultManager] requestImageDataForAsset:asset options:nil resultHandler:^(NSData * _Nullable imageData, NSString * _Nullable dataUTI, UIImageOrientation imageOrientation, NSDictionary * _Nullable info) {
+
+          if (imageData) {
+            // Get the original filename to keep it straight on the system.
+            // Alternatively could probably just use timestamp, but this doesn't seem to have any lag
+            NSArray *resources = [PHAssetResource assetResourcesForAsset:asset];
+            NSString *orgFilename = ((PHAssetResource*)resources[0]).originalFilename;
+
+            // Get our path in the tmp directory
+            NSString *path = [[NSTemporaryDirectory()stringByStandardizingPath] stringByAppendingPathComponent:orgFilename];
```
asutula: Nice. One thought/thing to keep an eye on... Depending on how long we need this image to exist for, we may want to consider writing the image to the documents dir. I'd imagine tmp has some sort of time where the OS deletes the file. Could be fine, though.

ios/NativeModules/TextileNode.m
andrewxhill: for sure... here you go https://github.com/textileio/textile-mobile/pull/535

PR #529, opened 2018-09-15
Tag Releases etc.

asutula:
Main changes here are:
- All merges to master upload adhoc builds to Crashlytics for internal testing
- Releases are triggered by tags like `[android,ios]_1.2.3_456`
- New tag-specified versions should be created by Fastlane if they don't exist in the App Store Connect and Google Play


asutula:
This guys is good to go

PR #533, opened 2018-09-18
Debug install beside production

andrewxhill:
screen shot 2018-09-17 at 5 12 40 pm
 
just changes this ^, but ended up going with WalletDebug so in fabric the nomenclature matched

allows side-by-side installs on your device

### DIFF
--- a/App/Services/DeepLink.ts
+++ b/App/Services/DeepLink.ts
@@ -2,6 +2,7 @@ import NavigationService from './NavigationService'
 import Config from 'react-native-config'
 import { ExternalInvite, DeepLinkData } from '../Models/TextileTypes'
 
+
 function getParams (hash: string): { [key: string]: (string | string[]) } {
   const query = hash.replace('#', '')
   const vars = query.split('&').map((expression) => expression.split('='))
@@ -39,7 +40,7 @@ function getData (href: string): DeepLinkData | undefined {
     host: match[2],
     hostname: match[3],
     port: match[4],
-    path: match[5],
+    path: match[5].replace('/debug', ''),
     search: match[6],
     hash: match[7]
   }
@@ -53,6 +54,9 @@ function createInviteLink (invite: ExternalInvite, threadName: string): string {
   hash.push(`name=${encodeURIComponent(threadName)}`)
   if (Config.TEMPORARY_REFERRAL) {
     hash.push(`referral=${encodeURIComponent(Config.TEMPORARY_REFERRAL)}`)
+  }
+  if (__DEV__) {
+    return `https://www.textile.photos/debug/invites/new#${hash.join('&')}`
   }
   return `https://www.textile.photos/invites/new#${hash.join('&')}`
 }
--- a/App/Services/DeepLink.ts
+++ b/App/Services/DeepLink.ts
@@ -2,6 +2,8 @@ import NavigationService from './NavigationService'
 import Config from 'react-native-config'
 import { ExternalInvite, DeepLinkData } from '../Models/TextileTypes'
 
+const DEBUG_PATH_PREFIX = '/debug'
+
 function getParams (hash: string): { [key: string]: (string | string[]) } {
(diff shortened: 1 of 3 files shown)


andrewxhill:
Okay, so to run this on your own device means we'll have to be able to sign the new bundle id as well

screen shot 2018-09-17 at 6 33 57 pm

can we add it to our provisioning profile for debug releases @asutula?

andrewxhill:
annnnnd there we go. 

screen shot 2018-09-17 at 7 41 03 pm

let me know if that works for you. tested on my phone, currently running two copies of Textile, so good

App/Services/DeepLink.ts
```diff
@@ -2,6 +2,8 @@ import NavigationService from './NavigationService'
 import Config from 'react-native-config'
 import { ExternalInvite, DeepLinkData } from '../Models/TextileTypes'
 
+const DEBUG_PATH_PREFIX = '/debug'
```
asutula: Would it be more simple to just support another url prefix, like `debug.textile.io`? Then we wouldn't have to support any special cases in the code like this.

ios/TextilePhotos.xcodeproj/project.pbxproj
```diff
@@ -1984,10 +1984,10 @@
 					"-ObjC",
 					"-lc++",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = io.textile.Wallet;
+				PRODUCT_BUNDLE_IDENTIFIER = io.textile.WalletDebug;
 				PRODUCT_NAME = TextilePhotos;
 				PROVISIONING_PROFILE = "2db150d1-101b-4ac0-b886-3e69f53beb8a";
-				PROVISIONING_PROFILE_SPECIFIER = "match Development io.textile.Wallet";
+				PROVISIONING_PROFILE_SPECIFIER = "match Development io.textile.WalletDebug";
```
asutula: If I understand this correctly, it changes the bundle id and provisioning profile for the main application target... So what happens when we do a release build? How do we use to real bundle id and provisioning profile?

App/Services/DeepLink.ts
andrewxhill: You'd still need to detect it in the code, yeah? How could that work different?

ios/TextilePhotos.xcodeproj/project.pbxproj
andrewxhill: Well, I'm changing the bundle id on Debug builds so that it is a duplicate app on your phone. So to do that, I needed a signing cert that was able to sign that new bundle id. Bundle id for the release build is still our original and uses the same signing cert we had before,

screen shot 2018-09-18 at 8 16 19 am

App/Services/DeepLink.ts
asutula: Haven't thought very deeply about this, but was imagining turning this into more of a configuration/environment thing. If the app is configured to create/process URLs of a specific prefix, that prefix being part of the .env or whatever, then everything should just work seamlessly and consistently, no special cases. It's related to a larger idea of having a debug/staging environment at all levels of the system we're building.

ios/TextilePhotos.xcodeproj/project.pbxproj
asutula: Not sure, I think that's what that SO is all about. Trying to learn more about the issue now...

andrewxhill: okay, sweet. let me know what you want to do

App/Services/DeepLink.ts
andrewxhill: we could do that really nicely by leveraging the .env (say .env.debug to standardize) files instead of this and still using that config in the deeplink.ts file. that way it would work seamlessly for both ios/android. i didn't do that this time around because i wasn't sure we had a rigid practice of .env yet.

PR #535, opened 2018-09-18
adds canDelete=false for android local photos

andrewxhill:


andrewxhill:
would be cool if someone wants to run android with verbose + storage options and go take some pictures to see if they are added correctly. in ios simulator and device had a difference, don't think there will be one on android, but cool to check

PR #536, opened 2018-09-18
Nav Adjustment

asutula:
This moves most of the thread/photo/comment related screens up into the PrimaryNavigation. This has a couple benefits:

1. No more hiding/showing the tab bar depending on the route... These screens are now pushed onto the stack that contains the tab bar (similar to the settings screen)
2. Allows us to directly push whichever screen is appropriate when a notification is tapped, and when a user navigates back from that screen, they are back at Notifications (Feed).
3. Sets us up for a new single image screen that can be displayed when certain notifications are tapped (have that in another branch at the moment)


App/Navigation/AppNavigation/Primary/TabNavigator/index.tsx
```diff
@@ -41,4 +41,9 @@ const nav = createBottomTabNavigator(
   }
 )
 
+nav.navigationOptions = {
+  // Hide the header from AppNavigator stack
+  header: null
```
asutula: When the tab navigator is displayed on the stack, hide the stack's nav bar because each of the tabs holds a stack navigator that displays a nav bar.

App/Navigation/AppNavigation/Primary/index.ts
```diff
@@ -1,18 +1,42 @@
...
+
+    AddFriends: ThreadsEditFriends,
+    ViewThread: ThreadDetail,
+    Comments,
+    ThreadSharePhoto: AddCaptionScreen,
+    WalletPicker
   },
   {
-    mode: 'card',
-    headerMode: 'none',
```
asutula: Now we do display a navigation header by default (tab controller hides it for itself)

App/Navigation/AppNavigation/Primary/index.ts
```diff
@@ -1,18 +1,42 @@
...
+    NotificationSettings,
+    Storage,
+    Mnemonic,
+    ChangeAvatar: UpdateAvatar,
+
+    AddFriends: ThreadsEditFriends,
+    ViewThread: ThreadDetail,
+    Comments,
+    ThreadSharePhoto: AddCaptionScreen,
+    WalletPicker
```
asutula: These screens all exist at this higher level now so we can just push any of them on whenever needed, really for notifications.

yarn.lock
```diff
@@ -435,7 +435,7 @@
     lodash "^4.2.0"
     to-fast-properties "^2.0.0"
 
-"@textile/go-mobile@^0.1.9":
+"@textile/go-mobile@0.1.9":
```
asutula: Not sure why this wasn't already updated, but this change happened when I ran `yarn install`

asutula: After rebasing on latest master

App/Sagas/NotificationsSagas.ts
```diff
@@ -81,8 +81,6 @@ export function * notificationView (action: ActionType<typeof NotificationsActio
         if (threadData) {
           yield put(PhotoViewingAction.viewThread(threadData.thread.id))
           yield put(PhotoViewingAction.viewPhoto(notification.data_id as PhotoId))
-          yield call(
-            NavigationService.navigate, 'ViewThread', { id: threadData.thread.id, name: threadData.thread.name })
```
asutula: Now we can just navigate straight to the comment, no need to push thread and then comment. I think once the single photo view is in, it might make more sense to navigate to it instead of comments.

PR #537, opened 2018-09-19
Single photo screen

### LINKED ISSUE #510: New single image screen
Use thread image card as contents for single image screen accessed from notifications

asutula:
Wired up to notifications screen.

G2G @channel

Fixes #510

### DIFF
--- a/App/Sagas/NotificationsSagas.ts
+++ b/App/Sagas/NotificationsSagas.ts
@@ -85,9 +85,18 @@ export function * notificationView (action: ActionType<typeof NotificationsActio
         }
         break
       }
-      case NotificationType.deviceAddedNotification:
       case NotificationType.photoAddedNotification:
-      case NotificationType.likeAddedNotification:
+      case NotificationType.likeAddedNotification: {
+        const threadData: ThreadData | undefined = yield select(threadDataByThreadId, notification.subject_id)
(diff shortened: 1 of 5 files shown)


App/Sagas/NotificationsSagas.ts
```diff
@@ -85,9 +85,18 @@ export function * notificationView (action: ActionType<typeof NotificationsActio
...
-      case NotificationType.deviceAddedNotification:
       case NotificationType.photoAddedNotification:
-      case NotificationType.likeAddedNotification:
+      case NotificationType.likeAddedNotification: {
+        const threadData: ThreadData | undefined = yield select(threadDataByThreadId, notification.subject_id)
+        yield call(TextileNode.readNotification, notification.id)
+        if (threadData) {
+          yield put(PhotoViewingAction.viewThread(threadData.thread.id))
+          yield put(PhotoViewingAction.viewPhoto(notification.data_id as PhotoId))
+          yield call(NavigationService.navigate, 'PhotoScreen')
```
asutula: So for photo adds and likes, we navigate right to the single photo screen. I'm also thinking that for photo comments, we could also navigate to single photo screen instead of comments screen, but be fore making that change, we should make `ThreadDetailCard` parameterizable in a way where it would be forced to display the comment that the notification is referring to. We can get to that in a later PR.

PR #540, opened 2018-09-20
adds hms verbose ui row

andrewxhill:


App/Containers/Styles/RootContainerStyles.js
```diff
@@ -26,8 +26,9 @@ export default StyleSheet.create({
     alignItems: 'center',
     position: 'absolute',
     width: '100%',
-    bottom: 0,
-    height: 20,
+    bottom: 85,
+    height: 22,
+    paddingTop: 4,
```
andrewxhill: moved it above the tab bar... in buttonless phones it was disappearing at the bottom. and on se it felt like it was in the way of my clicks.

PR #550, opened 2018-09-22
Start background task early in the foreground

asutula:
Maybe this will help @andrewxhill's phone not get suspended before we can process the transition to background.

### DIFF
--- a/App/Sagas/NodeLifecycle.ts
+++ b/App/Sagas/NodeLifecycle.ts
@@ -21,6 +21,8 @@ export function * manageNode () {
           action.type === getType(TextileNodeActions.appStateChange) && (action.payload.newState === 'active' || action.payload.newState === 'background')
         )
 
+      BackgroundTimer.start()
+
       if (yield select(PreferencesSelectors.verboseUi)) {
         yield call(displayNotification, 'App State Change: ' + action.payload.newState)
       }
@@ -43,7 +45,6 @@ export function * manageNode () {
       // Using the race effect, if we get a foreground event while we're waiting
       // to stop the node, cancel the stop and let it keep running
       if (action.payload.newState === 'background') {
-        BackgroundTimer.start()
         // This race cancels whichever effect looses the race, so a foreground event will cancel stopping the node
         yield race({
           delayAndStopNode: call(stopNodeAfterDelay, 20000),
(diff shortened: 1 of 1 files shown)


App/Sagas/NodeLifecycle.ts
```diff
@@ -21,6 +21,8 @@ export function * manageNode () {
           action.type === getType(TextileNodeActions.appStateChange) && (action.payload.newState === 'active' || action.payload.newState === 'background')
         )
 
+      BackgroundTimer.start()
```
andrewxhill: So this would start a timer even on a foreground event? Not sure I follow what the Timer does or what this change will do...

asutula: Yep. It starts the timer no matter what, even if you're getting a foreground event. It's actually the way Apple recommends to do it. They basically say "if you start something in the foreground that might need to keep running in the background, don't wait until you get an app transition to the background state to start the background task".

So, both these sequences are totally valid:

1. foreground event > start background task > start work > end work > end background task > background event
2. foreground event > start background task > start work > background event > end work > end background task

In our case, 'work' is running the node.

andrewxhill: Ahh, makes sense right. Like on my phone it kinda appeared as if it would sometimes be killed on the moment i'd background... this could be it

asutula: The idea with this change is to try to make sure that when your phone transitions to the background, we have definitely already started a background task. This will hopefully prevent you phone from suspending the app until after we've stopped the node.

asutula: Just as a comparison, the current/broken sequence is:

foreground event  > start work > background event > start background task> end work > end background task

asutula: And it seems that sometimes, the app gets suspended in between start work > background event or background event > start background task

asutula: Oh, and this npm module that provides the thing called `BackgroundTimer`... that's just what it calls the thing that wraps iOS's background task API. So `BackgroundTimer.start()` starts a background task, and ``BackgroundTimer.stop()` stops the background task. These calls always have to be balanced.

PR #553, opened 2018-09-23
Brute force approach

asutula:
Ok this one uses the fact that we can safely call create and start on the node even if is is already created and/or started. It's more of a brute force approach where we call create and start on the node for every background or foreground event we get no matter what. Additionally, if the event is a background event, we do our background task thing and let the node run for 20 seconds, but that can get interrupted by another foreground event.

@andrewxhill let's see how this works on that old phone of yours.


asutula:
If this doesn't work, I'm out of ideas.

andrewxhill:
now we're talking

PR #572, opened 2018-09-27
Fix JS error reporting

asutula:
Also updated Fastlane to deal with new iPhones.

Still need a workflow for stack symbolization, but this is a good first step.


App/Containers/ThreadsList.js
```diff
@@ -96,6 +96,8 @@ class ThreadsList extends React.PureComponent {
       })
     }
     if (
+      this.props.threads &&
+      prevProps.threads &&
```
asutula: This was a bug hiding if there are no threads to display and the logic in here says not to display the onboarding content.

App/Containers/App.tsx
```diff
@@ -49,6 +50,11 @@ class App extends Component {
     this.deepLinkEventHandler.tearDown()
     backgroundTaskEventHandler.tearDown()
   }
+
+  componentDidCatch(error: any, info: any) {
+    // TODO: Render some UI
+    errorHandler(error, false)
```
asutula: React swallows any errors that happen in `render` or any of the `Component` lifecycle methods and the unmounts the view hierarchy. This is probably what happens when you see a white screen in the app. This method is how you catch them and do something with them, so we should at least see those errors now.

andrewxhill: so will it crash the app now, or report the error but stay white screen?

asutula: Non fatal errors, which are the type that happen when a `Component` error happens, won't crash the app. The screen will just stay white. Buuuutt... Next thing is to render some sort of "Oops" screen when that type of error happens.

PR #577, opened 2018-10-01
Just cleans up the view

andrewxhill:


andrewxhill:
screen shot 2018-10-01 at 4 50 24 pm

just pulls in the standard header and back arrow so that the X series header doesn't break the while thing

andrewxhill:
yeah, we should just go with it. handles the header in the more standard way now too

PR #579, opened 2018-10-02
adding photo upload specific logging

andrewxhill:


andrewxhill:
Okay, this little loggin util is gtg if it would be useful in other cases. I didn't dump a ton of events into it, just the ones so i could find that image upload error

App/SB/views/DeviceLogs/index.js
```diff
@@ -0,0 +1,109 @@
...
+      </View>
+    )
+  }
+
+  render () {
+    return (
+      <View style={styles.container}>
+        <View style={styles.contentContainer}>
+          {this.renderHeader()}
+          <ScrollView>
```
asutula: You shouldn't need a `ScrollView` because a `FlatList` is a scroll view.

App/Redux/DeviceLogsRedux.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function reducer (state: DeviceLogsState = initialState, action: DeviceLogsAction): DeviceLogsState {
+  switch (action.type) {
+    case getType(actions.logNewEvent): {
+      const logRow: DeviceLogsRow = {
+        time: action.payload.time,
+        event: action.payload.event,
+        message: action.payload.message,
+        error: action.payload.error
+      }
```
asutula: Easier: `const logRow: DeviceLogsRow = { ...action.payload }`

App/Sagas/NodeLifecycle.ts
```diff
@@ -23,6 +24,7 @@ export function * manageNode () {
 
       if (yield select(PreferencesSelectors.verboseUi)) {
         yield call(displayNotification, 'App State Change: ' + action.payload.newState)
+        yield call(logNewEvent, new Date().getTime(), 'App State Change', action.payload.newState)
```
asutula: Should you move this outside the `if` here? Seems better that way so you don't miss some log data if you happen to turn off verbose UI.

App/Sagas/DeviceLogs.ts
```diff
@@ -0,0 +1,13 @@
+import {put, select} from 'redux-saga/effects'
+import DeviceLogsActions from '../Redux/DeviceLogsRedux'
+import {PreferencesSelectors} from '../Redux/PreferencesRedux'
+
+export function * logNewEvent (time: number, event: string, message: string, error?: boolean) {
+  if (yield select(PreferencesSelectors.verboseUi)) {
```
asutula: Same comment... Seems like we should always log so the data is there when you turn verbose UI on.

andrewxhill: intentional for now. i don't want to maintain any logs or deal with this flow on user's devices that aren't running verbose. let's go with it like this and if we see cases where it's really required without verbose ui we can sort it

asutula: Cool

PR #581, opened 2018-10-02
jpeg not heic

andrewxhill:


andrewxhill:
This fixes it, but was using the logging branch to get it all going. Going to go wrap that up and then this should be gtg

andrewxhill:
okay, this guy next! :)
it'll still be in trial mode for us, will be nice to see how it performs again

PR #582, opened 2018-10-02
Fix photo view/comments JS crash

asutula:


App/Containers/Comments.tsx
```diff
@@ -79,12 +79,9 @@ class Comments extends Component<Props> {
 
 const mapStateToProps = (state: RootState): StateProps  => {
   const { viewingPhoto } = state.photoViewing
-  if (!viewingPhoto) {
-    throw new Error('no viewing thread or photo')
-  }
```
asutula: In both these cases, I was throwing an exception here because I thought it should never happen. Bad idea because that crashes the JS and unmounts the entire UI.

Been digging into why `viewingPhoto` would ever be `undefined`... I can't figure it out, but these fixes should at least make it so the UI doesn't unmount.

asutula:
Still need to figure out what the underlying problem is with the redux state `viewingPhoto` going `undefined` for some reason. Where the UI used to go white, please keep an eye out for other strange behavior (some part of the UI going blank or flashing, for example).

asutula:
I checked to make sure the code that handles tapping a native notification and navigating to the photo view or comments view is intact... I looks good. Let's keep an eye on that once this fix is in.

App/Containers/Comments.tsx
carsonfarmer: Hmm, might this be a legacy issue? I think originally we didn't have comments at all, so maybe this was there to catch those cases?

asutula:
Na, it's code I wrote recently. Didn't feel good about it, but hoped it wouldn't be an issue.

PR #584, opened 2018-10-03
Micro delay before nav

andrewxhill:
Not sure why... probably worth a deeper look someday. But if you come out of a background state, notification clicks seem to dead end. But if I put this small delay in, they seem to navigate properly. 

That's probably why they work sometimes, the app is already warm when you click them


andrewxhill:
okay, let's give this a try. could need more investigation

PR #585, opened 2018-10-03
iPhone X fixes

asutula:
Use `SafeAreaView` in a few key places so padding is correct at top and bottom of the screen on iPhone Xish models:

1. Onboarding
2. Comments
3. Invite peers

There is also a layout issue for iPhone X with the sign up/in screens as well, but something wasn't straight forward about fixing it, so I bailed since we'll be getting rid of those screens soon with profiles coming.

@andrewxhill also mentioned a crash that happens in the sim sometimes when navigating back from the comments screen. Something about `Unsupported layout animation createConfig property (null)`. I tinkered around and realized that might also be a real crash for people that use hardware/bluetooth keyboards. This PR fixes that crash.

### DIFF
--- a/App/Components/KeyboardResponsiveContainer.tsx
+++ b/App/Components/KeyboardResponsiveContainer.tsx
@@ -26,30 +26,36 @@ export default class KeyboardResponsiveContainer extends React.Component<Props, 
   }
 
   keyboardWillChangeFrame = (event: any) => {
-    LayoutAnimation.configureNext(LayoutAnimation.create(
-      event.duration,
-      LayoutAnimation.Types[event.easing]
-    ))
+    LayoutAnimation.configureNext({
+      duration: event.duration,
+      update: {
+        type: LayoutAnimation.Types[event.easing]
+      }
+    })
     this.setState({
         height: this.bottomViewY - event.endCoordinates.screenY
     })
   }
 
   keyboardWillAppear = (event: any) => {
-    LayoutAnimation.configureNext(LayoutAnimation.create(
-      event.duration,
-      LayoutAnimation.Types[event.easing]
-    ))
+    LayoutAnimation.configureNext({
+      duration: event.duration,
+      update: {
+        type: LayoutAnimation.Types[event.easing]
+      }
+    })
     this.setState({
         height: this.bottomViewY - event.endCoordinates.screenY
     })
   }
 
   keyboardWillHide = (event: any) => {
-    LayoutAnimation.configureNext(LayoutAnimation.create(
-      event.duration,
-      LayoutAnimation.Types[event.easing]
-    ))
+    LayoutAnimation.configureNext({
+      duration: event.duration,
+      update: {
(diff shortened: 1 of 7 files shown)


asutula:
After digging into the iPhone X layout issues, I figured out the general rule of thumb is this: react-navigation handles iPhone X whenever there is a navigation header and/or tab bar displayed (it uses `SafeAreaView` internally), but you need to consider wrapping your component in a `SafeAreaView` any time:

1. Your component reaches the top of the screen and there is no navigation header displayed
2. Your component reaches the bottom of the screen and there is no tab bar displayed AND your component renders a "footer"-like view at the bottom (avoiding the bottom "swipe up bar" area doesn't seem to be an issue for normal content like a `FlatList` of things).

App/Containers/Comments.tsx
```diff
@@ -60,19 +60,21 @@ class Comments extends Component<Props> {
...
-        <ScrollView ref={(ref) => this.scrollView = ref ? ref : undefined} style={styles.contentContainer}>
-          <View>
-            {this.props.commentCardProps.map((commentCardProps, i) => (
-              <CommentCard key={i} {...commentCardProps} />
-            ))}
-          </View>
-        </ScrollView>
-        <CommentBox onUpdate={this.props.updateComment} onSubmit={this.props.submitComment} value={this.props.commentValue} />
-      </KeyboardResponsiveContainer>
+      <SafeAreaView style={styles.safeContainer}>
```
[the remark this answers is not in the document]
asutula: Yea. See my latest comment for the logic of when do do this.

App/Components/KeyboardResponsiveContainer.tsx
```diff
@@ -26,30 +26,36 @@ export default class KeyboardResponsiveContainer extends React.Component<Props,
...
+        type: LayoutAnimation.Types[event.easing]
+      }
+    })
     this.setState({
         height: this.bottomViewY - event.endCoordinates.screenY
     })
   }
 
   keyboardWillHide = (event: any) => {
-    LayoutAnimation.configureNext(LayoutAnimation.create(
```
asutula: This `LayoutAnimation.create` was creating `LayoutAnimation` configuration objects that were sometimes considered invalid and causing that crash. Just directly specifying the configuration object like this seems to fix the crash for some reason. Don't totally understand it, but cool!

App/SB/views/OnBoarding/index.js
```diff
@@ -26,29 +27,33 @@ export default class OnBoarding extends Component {
...
-        </View>
-        <Footer
-          currentPageIndex={step}
-          pages={pages}
-          onNext={this.onNext}
-          onSkip={this.onSubmit}
-          onSubmit={this.onSubmit}
-        />
+        <SafeAreaView style={styles.safeContainerTop} />
+        <SafeAreaView style={styles.safeContainerBottom}>
```
asutula: Little hack here: Use two `SafeAreaView`s to support different colors for the top and bottom parts of the screen.

PR #590, opened 2018-10-05
adds a background location prompt after some usage

andrewxhill:
realized we never ask people to enable it, so unless they are going into notification settings and finding it, they probably aren't trying it out. 

looking at the verbose ui, it definitely helps


asutula:
Would be nice to do these sorts of things in a saga, no reason it needs to be part of the UI. Next time I guess...

PR #591, opened 2018-10-05
Pulls the threads rendering out from the navigation view

andrewxhill:
the methods in there were getting a bit wild. splits it out into the container and a component


App/Containers/ThreadsList.js
```diff
@@ -132,40 +132,6 @@ class ThreadsList extends React.PureComponent {
...
-          keyExtractor={this._keyExtractor}
-          renderItem={this._renderItem}
-          refreshing={this.props.refreshing}
-          onRefresh={this._onRefresh}
-          initialNumToRender={4}
-        />
-      </View>
-    )
-  }
-
```
asutula: Really not much left in this class. Seems like you just moved 90% of the code to another file. Maybe I'm missing something.

asutula: I guess navigation and placeholder stuff is handled in here... ok ok

PR #595, opened 2018-10-08
Full sharing flow

andrewxhill:
screens

Lots of UI work to be done now, but the flow is working. Feels about right...


sanderpick:
Slick! Looks way nicer. Related thoughts... 

- Could list contacts under the thread list in your flow ^... pick one and it creates the thread (or uses existing 1-1)
- From wallet view, select 1 or more photos, landing you in your flow... add caption to each, choose thread(s) / contact(s)

andrewxhill:
Totally, the first bullet was the motivation for the whole thing. I was thinking how 1to1 sharing was a bit broken in the old thread flow. So this should be pretty cool now. 

Also, gives a really nice 'home' for receiving a 1v1 share, which is just your full stream. You could drill into a Contact and then see just photos shared from that contact (but they could be mixed 1v1 and other threads they are part of where they shared with you). 

Anyway. Just pushed cleanup so this all works.

Next step I think is to resort the onboarding text stuff.

andrewxhill:
Oh... another benefit. We can make that "comment / pick thread or user" view much nicer and it fills the same role we'll need in the Native Share dialog

andrewxhill:
Okay, since this was still open I went ahead and pushed updates here

* Onboarding screens are fixed to match
* You now can create a thread even while just adding a photo to the main feed, see:

screens

(i removed that list sort you see though)

asutula:
@andrewxhill looks like you need to update a snapshot. Want to do that? Or let me know if you want me to.

andrewxhill:
oop fixed

PR #598, opened 2018-10-09
Run photo sync on every foreground, backgroud trigger, and location trigger

andrewxhill:


App/Sagas/NodeLifecycle.ts
```diff
@@ -37,6 +38,9 @@ export function * manageNode () {
       // or by launching into the background because of a trigger.
       if (action.payload.newState === 'background') {
         yield fork(backgroundTaskRace)
+      } else {
+        // Check for new photos on every Foreground event
+        yield put(StorageActions.refreshLocalImagesRequest())
```
asutula: This will work most of the time, but it's not guaranteed to work. Refreshing the local images eventually relies on the node being created/started/online, but you're refreshing the local images here without any guarantee that the node will be ready.

App/Sagas/NodeLifecycle.ts
```diff
@@ -111,10 +115,14 @@ function * stopNodeAfterDelay (ms: number) {
 }
 
 export function * backgroundTask () {
+  // Check for new photos on every background task event
+  yield put(StorageActions.refreshLocalImagesRequest())
```
asutula: Same goes for here and below.

asutula: So doing it this way, we have to rely on the processing of the photo query results to be resilient enough to do the right thing in the case of not being able to interact with the node. Not sure if that's the case or not, but something to keep in mind.

PR #605, opened 2018-10-09
Fix comment click

andrewxhill:
cleans up an issue i saw clicking 'comments' from the joined feed, since it didn't ever register what thread.

also fixes feed link clicks which needed similar


App/Sagas/NodeLifecycle.ts
```diff
@@ -99,6 +96,8 @@ function * stopNodeAfterDelay (ms: number) {
       if (yield select(PreferencesSelectors.verboseUi)) {
         yield call(displayNotification, 'Delayed stop of node canceled because of foreground event')
       }
+      // Check for new photos in case user left app and came back after taking one
+      yield put(StorageActions.refreshLocalImagesRequest())
```
asutula: Nice

PR #613, opened 2018-10-11
Tag release

andrewxhill:


andrewxhill:
cool, so this is GTG

.circleci/config.yml
```diff
@@ -242,6 +241,23 @@ jobs:
...
+        working_directory: ~/textile-mobile
+        docker:
+            - image: cibuilds/github:0.10
+        steps:
+            - attach_workspace:
+                at: ~/textile-mobile
+            - deploy:
+                name: release apk
+                command: |
+                    ghr -t ${GITHUB_TOKEN} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -c ${CIRCLE_SHA1} -delete ${CIRCLE_TAG} android/app/build/outputs/apk/*.apk
```
asutula: So what does this do exactly? Specifically curious about the `-delete` option.

andrewxhill: ah shoot, good call meant to drop that

asutula: But also just curious about what happens... Like, does this just add the apk as an attachment to the existing release on GH? (because GH creates a 'release' when you create the tag that triggers this build)

andrewxhill: right so, ${CIRCLE_TAG} is going to be the tag associated with that Release. This will upload the artifacts there, so yep existing. --delete would have recreated the release which we don't want, we just want to upload it to the exiting.

andrewxhill: you can see the example one created before i delete it, https://github.com/textileio/textile-mobile/releases/tag/test_apk and the tag it pushed to https://github.com/textileio/textile-mobile/tags

asutula: cool perfect

asutula: now i just have to get the android build working again hahah!

PR #616, opened 2018-10-12
Remove react-native-background-task

asutula:
and use `react-native-background-fetch` directly.

`react-native-background-task` is causing android build issues, hasn't been updated in a year, and seems to not add any functionality on top of `react-native-background-fetch`.


App/Redux/TriggersRedux.ts
```diff
@@ -1,7 +1,7 @@
 import { createAction, ActionType, getType } from 'typesafe-actions'
 
 const actions = {
-  backgroundTask: createAction('BACKGROUND_TASK'),
+  backgroundFetch: createAction('BACKGROUND_FETCH'),
```
asutula: Updated to 'fetch' working just to make things more clear.

typings/react-native-background-fetch/index.d.ts
```diff
@@ -0,0 +1,31 @@
+declare module 'react-native-background-fetch' {
```
asutula: Wrote the typings here. Could contribute to DefinitelyTyped, need to look into that.

asutula:
Ok gonna merge this guy. It hopefully gets the Android build working again, and if not, I want to keep going on fixing it.

PR #634, opened 2018-10-17
Nominal Types

asutula:


asutula:
This is good to go @andrewxhill

App/Components/ProgressiveImage.tsx
```diff
@@ -2,9 +2,10 @@ import React, { Fragment } from 'react'
 import { View } from 'react-native'
 import { ImageStyle, Platform } from 'react-native'
 import TextileImage from '../../TextileImage'
+import { PhotoId } from '../Models/TextileTypes'
 
 export interface IProgressiveImageProps {
-  imageId: string,
+  imageId: PhotoId,
```
asutula: Decided that UI components that really require a certain id type should just be typed strongly.

asutula: In contrast, if we create some general purpose/reusable UI component, it's props should just be normal types like `string` or `number`, for example.

App/Services/Notifications.ts
```diff
@@ -9,23 +9,65 @@ export interface INotificationsPayload {
   typeString: string,
 }
 
-export function isPhoto(notification: Notification): boolean {
-  if (!notification.data_id) {
-    return false
-  }
-  switch (notification.type) {
-    case NotificationType.receivedInviteNotification:
+export function toTypedNotification(notificationData: NotificationData): Notification {
```
asutula: This is the function that converts the untyped `NotificationData` to type safe notifications.

App/Services/Notifications.ts
```diff
@@ -9,23 +9,65 @@ export interface INotificationsPayload {
...
+  const { subject_id, subject, data_id, type, ...baseNotification } = notificationData
+  switch (type) {
+    case NotificationType.commentAddedNotification:
+      return {
+        ...baseNotification,
+        type,
+        threadId: notificationData.subject_id as any,
```
asutula: Any time you have a trusted normal value (`string` in this case) that you need to convert to the nominal type, you use `as any`.

App/Models/TextileTypes.ts
```diff
@@ -5,35 +5,21 @@
...
+export type DeviceId = string & { _deviceIdBrand: void }
+export type ProfileAvatarId = string & { _profileAvatarIdBrand: void }
+export type NotificationId = string & { _notificationIdBrand: void }
+
+export type UserName = string & { _userNameBrand: void }
+export type ThreadName = string & { _threadNameBrand: void }
+export type DeviceName = string & { _deviceNameBrand: void }
+export type PublicKey = string & { _publicKeyBrand: void }
+export type PrivateKey = string & { _privateKeyBrand: void }
+export type Mnemonic = string & { _mnemonicBrand: void }
```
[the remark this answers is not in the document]
andrewxhill: nice

PR #658, opened 2018-10-22
.env based builds

asutula:


asutula:
Ok G2G. 

Tested dev and adhoc builds thoroughly, release should work, but will fix any problems we encounter when we get there.

This allows 3 versions of the app to be installed on your phone... Dev, Beta, and Textile for local Xcode builds, Crashlytics adhoc distributions, and App Store releases, respectively. 

Everyone should have gotten invite emails for the new app ids.

The Xcode project is now configured to build the 'dev' version of the app, so if you regularly build it locally, you'll want to install the new provisioning profiles. The the `ios` folder:
```
fastlane match development -a io.textile.Wallet.dev
fastlane match adhoc -a io.textile.Wallet.dev
```
For that to work, you'll need our `.env.default` in the `ios` folder... Let me know if you have any issues there.

PR #659, opened 2018-10-22
Wallet freshen up

andrewxhill:


App/Containers/Wallet.js
```diff
@@ -19,11 +23,16 @@ import { RootState } from '../Redux/Types'
 class Wallet extends React.PureComponent {
   static navigationOptions = ({ navigation }) => {
     const params = navigation.state.params || {}
-    const headerTitle = params.username ? 'Hello, ' + params.username : 'Hi there!'
+
+    const headerRight = (
+      <TextileHeaderButtons>
+        <Item title='Settings' iconName='cog' onPress={params.updateSettings}/>
```
andrewxhill: This is what i need for now @asutula

asutula: Cool, can make that happen

PR #668, opened 2018-10-25
Nav to invited thread

asutula:
Simplified/TS-ified thread invite screen, nav to thread once it is joined and added


App/Sagas/DeepLinkSagas.ts
```diff
@@ -24,7 +24,7 @@ export function * inviteAfterOnboard () {
   if (invite) {
     // ensures this is the last of the knock-on effects of onboarding
     yield call(delay, 250)
-    NavigationService.navigate('ThreadInvite', { url: invite.url, request: DeepLink.getParams(invite.hash) })
```
asutula: `url` wasn't used anywhere as far as i can see

App/Services/DeepLink.ts
```diff
@@ -65,7 +65,7 @@ function route (link: string) {
       NavigationService.navigate('PairingView', { request: getParams(data.hash) })
     } else if (data.path === '/invites/new' && data.hash !== '') {
       // invite the user to the thread
-      NavigationService.navigate('ThreadInvite', { link, request: getParams(data.hash) })
```
asutula: `link` wasn't used anywhere as far as i can see

asutula:
This seems to work ok. I don't love that it immediately navigates away from the accept invite screen and a little time may pass before it navigates to the thread. I think that can be cleaned up with some blocking UI and possibly a different navigation setup for thread invites. Maybe this is a good first step and we can improve all that.

PR #671, opened 2018-10-25
Parameterize url scheme

asutula:
Defaults to `textile-dev://`, updated in env specific builds on CircleCI. Schemes are as follows:
 
- production: `textile`
- beta: `textile-beta`
- dev: `textile-dev`

So, when you have an invite URL, you can choose with version of the app to open it with by editing the URL scheme accordingly.

Made an update where you can have a url with or without `www.` in it (important because our web page redirects to `www.textile.photos` then you may copy that url.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -559,14 +559,12 @@ workflows:
                     branches:
                         only:
                             - master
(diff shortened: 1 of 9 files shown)


asutula:
Ok all good here.

PR #672, opened 2018-10-25
Moving Threads to a reusable modal

andrewxhill:
Primary goal was to help with the Nav mess of having it in the view stack

Ended up allowing me to clean up a few other parts of the UX

Don't love the bottom slide-up look yet, but ideally will replace the modal with a full screen swiss army knife that allows you to select contacts to add at the same time you create the thread

already lots of changes though, so breaking it off here and will continue as smaller parts


App/Navigation/AppNavigation/Primary/TabNavigator/index.tsx
```diff
@@ -37,7 +37,7 @@ const nav = createBottomTabNavigator(
     },
     animationEnabled: false,
     swipeEnabled: false,
-    initialRouteName: 'Wallet'
+    initialRouteName: 'Threads'
```
asutula: Just making sure you want this in there? Thought you wanted to highlight Wallet.

andrewxhill:
okay, gtg

App/Containers/Wallet.js
```diff
@@ -29,21 +26,18 @@ class Wallet extends React.PureComponent {
...
-    const accountBackupEnabled = params.storage && params.storage.enableWalletBackup.status
-    const pinEnabled = params.storage && params.storage.autoPinPhotos.status
     const photoBackupEnabled = params.storage && params.storage.enablePhotoBackup.status
 
     const headerLeft = (
       <TextileHeaderButtons>
         <Item
           title='backup'
-          iconName='cloud'
+          iconName={photoBackupEnabled ? 'cloud-checked' : 'cloud'}
```
asutula: Nice. You can also set the color, just pass a `color` prop set to brandBlue or whatever. If you want.

andrewxhill: actually didn't work... didn't dig deep into it but yeah, color param was there before and not working

asutula:
Ok. Good enough for now

PR #673, opened 2018-10-25
New icon set, improved icon colors

asutula:
The goal here is mainly to use a good enough icon set for now that generally has everything we need instead of spending time creating shitty icons myself. So, you'll notice some icons you'll like more than the old ones, and some you'll like less. Overall, it's an improvement, though. 

**They will initially look bad to you because you're used to the old ones... Give it a little time.**

Another improvement is making the icons in the header the same color as in the tab bar. Equalizes their importance a bit.


asutula:
screen shot 2018-10-25 at 18 35 46
Demo.

asutula:
Also, I just picked the icons I thought were appropriate, let me know if any of them need to change.

asutula:
[icomoon(1).zip](https://github.com/textileio/textile-mobile/files/2517114/icomoon.1.zip)
Unzip this and open `demo.html` if you want to browse all the available icons in this set

PR #677, opened 2018-10-26
Header button color prop support

asutula:
Fixed the bug


andrewxhill:
can't recall, it might have been that way to disable 'Share' text in the header

asutula:
Either way, that `color` prop is used to set both the icon and text color.

PR #681, opened 2018-10-26
Placeholder in peers tab

andrewxhill:
Tired of looking at the empty peers tab. Placeholder, will swap in real contacts api for data when merged


andrewxhill:
should be gtg

PR #682, opened 2018-10-28
Use minWidth api instead of path api

andrewxhill:
Trying to update this API hitting a snag.

```
        <TextileImage
          imageId={this.props.imageId}
          minWidth={this.props.minWidth}
          style={[{ backgroundColor: 'red' }, this.props.style || {}]}
          resizeMode={resizeMode}
          capInsets={this.props.capInsets}
        />
```

okay, red, you'll see why in a sec. capinsets is undefined. ```resizeMode={'contain'}```, we'll come back to why not `cover` last. style is 

```
{ alignItems: 'center',
       padding: 0,
       margin: 0,
       width: 375,
       height: 562.0503597122303 }
```

So... then in TextileImage, I've been trying to use `PixelRatio.getPixelSizeForLayoutSize(this.props.minWidth)` so that we could show true res images based on the screen. Weird things happen though.

Doing it like this,

```
    const nativeProps = {
      ...this.props,
      minWidth: this.props.minWidth,
      onLoad: this._onLoaded.bind(this),
      onError: this._onError.bind(this)
    }
    return <TextileImageView {...nativeProps} />
```

results in

screen shot 2018-10-28 at 12 07 30 pm

This more or less emulates what we used to have. All good.

Trying this though,

```
    const nativeProps = {
      ...this.props,
      minWidth: Math.floor(this.props.minWidth * 1.2),
      onLoad: this._onLoaded.bind(this),
      onError: this._onError.bind(this)
    }
    return <TextileImageView {...nativeProps} />
```

Will get you,

screen shot 2018-10-28 at 12 10 19 pm

I don't get how that background padding exists... and based on the width it will get worse. 

Take a look now if I go to full computed pixels,

```
    const nativeProps = {
      ...this.props,
      minWidth: PixelRatio.getPixelSizeForLayoutSize(this.props.minWidth),
      onLoad: this._onLoaded.bind(this),
      onError: this._onError.bind(this)
    }
    return <TextileImageView {...nativeProps} />
```

You get,

screen shot 2018-10-28 at 12 11 32 pm

Fully off the screen.

Need to figure out where I'm going wrong in this rendering.

Caveat, setting resizeMode = cover we get a new problem. Here I set to retain and use that same full pixel calculation. It'll be like,

screen shot 2018-10-28 at 12 13 18 pm


andrewxhill:
Okay, this is gtg.

Fix, don't use parameter name, `minWidth`

App/SB/components/ThreadDetailCard/index.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Dispatch } from 'redux'
 import { connect } from 'react-redux'
-import { View, Text, Dimensions, TouchableOpacity, ImageURISource, ImageStyle } from 'react-native'
+import { View, Text, Dimensions, TouchableOpacity, ImageURISource, ImageStyle, PixelRatio } from 'react-native'
```
asutula: nit... `PixelRatio` unused

andrewxhill:
1. Never ran into a case of it. Since we had this problem before, of not being able to render images properly without knowing the aspect ratio, we also always know the width.
2. I could do this, let me know which you prefer. Doesn't feel right to me on first think. Just because there are valid cases where you want the image to be rendered in preparation for a different final width. Example, animations/transitions.

asutula:
Right, that's why it should be optional. You can always force a `minWdith` by providing it regardless of `width`, if not provided, use `width`, if no `width`, use some default.

PR #687, opened 2018-10-30
Update to go-mobile 1.0

asutula:
G2G


App/Containers/StatusCheck.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+interface StateProps {
+  onboarded: boolean
+  nodeStarted: boolean
+  nodeError?: string
+}
+
+type Props = StateProps & NavigationScreenProps<{}>
+
+class StatusCheck extends React.Component<Props, {}> {
```
asutula: This is basically what `OnboardingCheck` was, just using it as a general use hub for checking state and making decisions about where to navigate. This adds the option of loading the migration flow.

android/app/src/main/java/com/textile/textilenode/TextileImageViewManager.java
```diff
@@ -34,7 +34,7 @@ public TextileImageView createViewInstance(ThemedReactContext context) {
 
     @ReactProp(name = "imageId")
     public void setImageId(TextileImageView textileImageView, String imageId) {
-        textileImageView.setImageId(imageId);
+        textileImageView.setImagePath(imageId + "/0");
```
asutula: @sanderpick I'm going to deal with that creating of the "path" to pass to `imageFileDataForMinWidth` here in the native code. The JS-facing property is simply `imageId`, and this code appends `/0` to it. That's a safe assumption to make for now, right?

ios/NativeViews/TextileImageViewManager.m
```diff
@@ -16,7 +16,10 @@ - (UIView *)view
   return [[TextileImageView alloc] initWithBridge:self.bridge];
 }
 
-RCT_EXPORT_VIEW_PROPERTY(imageId, NSString)
+RCT_CUSTOM_VIEW_PROPERTY(imageId, NSString, TextileImageView) {
+  NSString *imagePath = json ? [NSString stringWithFormat:@"%@/0", json] : defaultView.imagePath;
+  view.imagePath = imagePath;
```
asutula: Same thing for iOS here.

android/app/src/main/java/com/textile/textilenode/TextileImageViewManager.java
sanderpick: You don't want to make it generic so it can traverse other types of files from other schemas?

sanderpick: I was imagining you'd leverage `index` from the file object to create the path. Like: `<ThreadFilesInfo.target>/<ThreadFileInfo.index>`

asutula: Oh I see. Yea we can do that. So `TextileImage` would have `target` and `index` props on it in the JS layer?

asutula: Done.

asutula:
Biggest PR so far???

We should start testing this internally ASAP. It's too big to understand. Just use it and log the bugs lol.

PR #688, opened 2018-10-30
iOS pinch zoom

andrewxhill:
dabbled in android and it was a bit of pita, see that here https://kmagiera.github.io/react-native-gesture-handler/docs/getting-started.html#with-wix-react-native-navigation-https-githubcom-wix-react-native-navigation


andrewxhill:
Nothing on android yet, just same old image. It'll just take a small bit of work to integrate the library in android, but figured since none of us use it, we could test the quality of this in just ios first.

andrewxhill:
we should give this a go internally... make sure pinch zoom doesn't lead to boggy ui etc. let me know if gtg

PR #705, opened 2018-11-09
modals tested on large and small ios screens

andrewxhill:
carson, this could use a look on android when you have some minutes


andrewxhill:
you have a few to try this out on android carson?

PR #713, opened 2018-11-28
making the radio button a touch target for the thread select screen

saranrapjs:
textile is great! i've honestly waiting for something like this to bubble up and am excited to see how usable it is even for not p2p heads like me :)
this PR fixes something i ran into when initially poking around in the app, where tapping on the radio button in the "Thread select" view for an image wouldn't select the corresponding thread. hopefully uncontroversial, but let me know if there's anything i should do differently

### DIFF
--- a/App/SB/components/RadioButton/index.js
+++ b/App/SB/components/RadioButton/index.js
@@ -1,21 +1,17 @@
 import React from 'react'
-import { View, TouchableOpacity, Image } from 'react-native'
+import { View, Image } from 'react-native'
 
 import styles from './statics/styles'
 
 const RadioButton = props => {
-  const { selected, style, disabled, onPress } = props
+  const { selected, style, disabled } = props
 
   return (
-    <TouchableOpacity style={[styles.button, style, disabled && styles.disabled]} onPress={onPress}>
+    <View style={[styles.button, style, disabled && styles.disabled]}>
       { selected && <Image style={styles.buttonImage} source={require('./statics/check.png')} /> }
       { selected && <View style={[styles.buttonSelected, disabled && styles.buttonDisabled]} /> }
-    </TouchableOpacity>
+    </View>
   )
-}
-
-RadioButton.defaultProps = {
-	onPress: () => {},
 }
 
 export default RadioButton
--- a/App/SB/components/RadioButton/index.js
+++ b/App/SB/components/RadioButton/index.js
@@ -4,14 +4,18 @@ import { View, TouchableOpacity, Image } from 'react-native'
 import styles from './statics/styles'
 
 const RadioButton = props => {
-  const { selected, style, disabled } = props
+  const { selected, style, disabled, onPress } = props
 
   return (
-    <TouchableOpacity style={[styles.button, style, disabled && styles.disabled]}>
+    <TouchableOpacity style={[styles.button, style, disabled && styles.disabled]} onPress={onPress}>
       { selected && <Image style={styles.buttonImage} source={require('./statics/check.png')} /> }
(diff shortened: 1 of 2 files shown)


asutula:
Awesome, thanks for the PR @saranrapjs! Looking at the code now, I think there's a (more) simple fix. Will insert a couple comments.

App/SB/components/RadioButton/index.js
```diff
@@ -4,14 +4,18 @@ import { View, TouchableOpacity, Image } from 'react-native'
...
   return (
-    <TouchableOpacity style={[styles.button, style, disabled && styles.disabled]}>
+    <TouchableOpacity style={[styles.button, style, disabled && styles.disabled]} onPress={onPress}>
       { selected && <Image style={styles.buttonImage} source={require('./statics/check.png')} /> }
       { selected && <View style={[styles.buttonSelected, disabled && styles.buttonDisabled]} /> }
     </TouchableOpacity>
   )
 }
 
+RadioButton.defaultProps = {
```
asutula: First, this component is just a simple stateless component. It's a function that takes in props and return a component. Setting `defaultProps` has no effect I think.

App/SB/components/RadioButton/index.js
```diff
@@ -4,14 +4,18 @@ import { View, TouchableOpacity, Image } from 'react-native'
 import styles from './statics/styles'
 
 const RadioButton = props => {
-  const { selected, style, disabled } = props
+  const { selected, style, disabled, onPress } = props
 
   return (
-    <TouchableOpacity style={[styles.button, style, disabled && styles.disabled]}>
```
asutula: This original `TouchableOpacity` had no function. It's use was actually causing the bug I understand you're trying to fix. This `TouchableOpacity` would "steal" the touch event from the `TouchableOpacity` in the parent component (the one that actually has the touch handler callback on it), and so if you touch the radio button, nothing happens.

The simple solution is to just change the use of `TouchableOpacity` here to a simple `View` since the `TouchableOpacity` contributed nothing anyway. This will allow the touch handler in the parent component to be triggered if you touch anywhere in the row, including the radio button.

App/SB/components/ThreadSelect/ThreadSelectCard.tsx
```diff
@@ -25,21 +25,23 @@ interface ScreenProps {
...
-          if (this.props.onSelect) {
-            this.props.onSelect(this.props.thread.id)
-          }
-        }}
+        onPress={this._onPress.bind(this)}
       >
         <PhotoWithTextBox text={this.props.thread.name} photo={this.props.thumb} />
         <View style={styles.threadSelectRadio}>
-          <RadioButton selected={this.props.selected} disabled={this.props.disabled}/>
+          <RadioButton selected={this.props.selected} disabled={this.props.disabled} onPress={this._onPress.bind(this)} />
```
asutula: All the changes in this file couple be reverted if you do that.

App/SB/components/RadioButton/index.js
saranrapjs: Ah, roger that! This is the first time I'm working with a react-native project, so I wasn't familiar with the distinction. I've reverted everything but the change in component type

App/SB/components/RadioButton/index.js
saranrapjs: Stateless components are no different from class components: they will pull in static properties like `defaultProps`/`propTypes` in the same way. I wasn't exactly sure where all else this radio button was being used, & had wanted to have a default in case!

asutula: Wow cool, didn't know!

PR #718, opened 2018-12-04
moves a bunch of repeated logic out of the maptoprops

andrewxhill:


andrewxhill:
was originally just looking for places where we were doing the filter by thread.name === default and ended up just trying to consolidate a bunch of different thread transforms.

App/Components/ThreadSelector.tsx
```diff
@@ -25,8 +25,26 @@ class ThreadSelector extends React.Component<ScreenProps & DispatchProps & Navig
...
+      updated: thread.photos.length > 0 && thread.photos[0].date ? Date.parse(thread.photos[0].date) : 0,
+      // latest peer to push to the thread
+      latestPeerId: thread.photos.length > 0 && thread.photos[0].author_id ? thread.photos[0].author_id : undefined
+    }
+  }
+
   _renderItem = (rowData: any) => {
-    const item: ThreadData = rowData.item
+    const thread: ThreadData = rowData.item
+    const item = this._convertToPreview(thread)
```
asutula: This conversion should happen in mapStateToProps I think. And the conversion function could effectively be a "selector" and just live with the rest of the selector code.

App/Components/ContactModal.tsx
```diff
@@ -75,26 +74,12 @@ class ContactModal extends React.Component<DispatchStateProps & ScreenProps> {
...
-          id: thread.id,
-          thumb: thread.photos.length > 0 ? thread.photos[0] : undefined,
-          name: thread.name
-        }
-      })
-  }
-
   return {
-    threads: threads || []
+    threadThumbs: getThreadThumbs(state, ownProps.peerId)
```
asutula: awesoeme

App/Containers/ThreadsList.js
```diff
@@ -216,9 +194,9 @@ const mapStateToProps = (state) => {
   return {
     profile,
     items,
-    showNotificationsPrompt: state.preferences.tourScreens.notifications && threads,
+    showNotificationsPrompt: state.preferences.tourScreens.notifications && items.length,
```
asutula: Just curious about JS truthy stuff like this... so `&& items.length` is `true` if `length` > 0? Or what?

asutula: But noticed below you actually compare with `=== 0`, so just want to be sure this was intentional.

asutula:
Related to this... I think we could use selectors way more in general, keep our mapStateToProps more simple and efficient. There's a great library we should look at the makes selectors as efficient as possible: https://github.com/reduxjs/reselect. Not something to worry about today, but in the future.

App/Containers/ThreadsList.js
andrewxhill: yeah, items.length > 0 => true.

i'm not certain this is going to behave perfectly, there are a few cases of this to ensure tour screens aren't shown just in the time before redux has rehydrated... would be nice to know the universal approach to doing that.

App/Components/ThreadSelector.tsx
andrewxhill: done

App/Redux/PhotoViewingSelectors.ts
```diff
@@ -20,9 +20,73 @@ export function photoAndComment (state: RootState) {
   }
 }
 
-export function getThreads (state: RootState): ReadonlyArray<ThreadData> {
+// TODO: replace with contacts api
```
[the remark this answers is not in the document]
andrewxhill: yeah, was just isolating it to do as a later PR

App/Redux/PhotoViewingSelectors.ts
```diff
@@ -20,9 +20,73 @@ export function photoAndComment (state: RootState) {
...
+        }
+        if (photo.author_id) {
+          peers[photo.author_id] = peers[photo.author_id] || photo.username || ''
+        }
+      }
+    })
+  return peers
+}
+
+export function getThreads (state: RootState, sortBy?: string): ReadonlyArray<ThreadData> {
```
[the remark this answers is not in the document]
andrewxhill: cool, done

andrewxhill:
sweet, should be gtg

PR #719, opened 2018-12-05
Filter Account thread from displayed photo threads

andrewxhill:


App/Redux/AccountSelectors.ts
```diff
@@ -0,0 +1,13 @@
+import { RootState } from './Types'
+
+/**
+ * Returns the Account Address
+ * Used to filter Account Thread out of Thread lists for display
+ * Account Thread Key === Account Address
+ */
+export function getAddress (state: RootState): string | undefined {
+  return state.account &&
+         state.account.profile &&
```
asutula: These first two above are always defined, you can remove them.

andrewxhill: odd, i got an error there, but might have been something else i was doing, moved this around a few times

andrewxhill:
k, gtg

PR #725, opened 2018-12-06
first

andrewxhill:
General cleanup of invite component

Fixes crash coming from a fork call by replacing with a call


andrewxhill:
okay, that should be gtg

andrewxhill:
oh nice, didn't see that the target was back up to lint speed. done and done

PR #733, opened 2018-12-12
Better handling of image upload complete and retry

### LINKED ISSUE #728: Photos being added to threads multiple times sometimes
Think this is because of sloppy state checking when uploads complete. There is a chance that each completed upload detects that it needs to share to the thread. There are three uploads per shared file right now.

### LINKED ISSUE #734: Verify session refresh retry of uploads

asutula:
Async is hard. Now that we have multiple images that need to be uploaded for a single share, those uploads are harder to manage. I think this does the trick and fixes the bug in the latest release where you'd see multiple shares of a single image that should have just been shared once. 

Updated to also handle session management and upload retry.

Fixes #728 
Fixes #734


App/Redux/ProcessingImagesRedux.ts
```diff
@@ -202,6 +208,20 @@ export function reducer (state: ProcessingImagesState = initialState, action: Pr
       })
       return { ...state, images }
     }
+    case getType(actions.uploadError): {
```
asutula: We need to track errors for individual uploads now, this special action is for that. The other `error` action is what's used for higher level error around preparing the image, adding it to a thread, etc.

App/Sagas/ImageSharingSagas.ts
```diff
@@ -145,6 +148,17 @@ export function * uploadPins (uuid: string) {
   }
 }
 
+export function * monitorForUploadsComplete(uuid: string) {
```
asutula: This is cool. Once the uploads are started for a single share, this saga waits for them all to finish before calling `shareToThread`. Uses a special `waitFor` effect that I wrote.

App/Sagas/ImageSharingTriggers.ts
```diff
@@ -37,11 +38,13 @@ export function * handleImageUploadComplete (action: ActionType<typeof Processin
...
     } catch (e) {}
-    const allComplete: boolean = yield select(allUploadsComplete, processingImage.uuid)
-    const alreadySharing = processingImage.status === 'sharing' || processingImage.status === 'complete'
-    if (allComplete && !alreadySharing) {
-      yield call(shareToThread, processingImage.uuid)
-    }
+  }
+}
+
+export function * startMonitoringExistingUploads () {
```
asutula: If the app is booted up while there are uploads in flight, this saga kicks off the monitoring of those in flight uploads so we'll know when they've completed and move on to sharing the image.

App/Sagas/WaitFor.ts
```diff
@@ -1,14 +1,14 @@
-import { select, take } from 'redux-saga/effects'
+import { select, take, SelectEffect } from 'redux-saga/effects'
 import { RootState } from '../Redux/Types'
 
-export function * waitFor (selector: (state: RootState) => boolean) {
-  if (yield select(selector)) {
-    return
+export function * waitFor (selectEffect: SelectEffect) {
```
asutula: This is a reusable effect I created that makes it easy to block a saga until some state you are interested in exists in the redux store.

App/Services/EventHandlers/UploadEventHandler.ts
```diff
@@ -26,16 +26,16 @@ export default class UploadEventHandler {
     } else if (responseCode === 401) {
       this.store.dispatch(ProcessingImagesActions.expiredTokenError(e.id))
     } else {
-      this.store.dispatch(ProcessingImagesActions.error(e.id, `Response code: ${responseCode}`))
+      this.store.dispatch(ProcessingImagesActions.uploadError(e.id, `Response code: ${responseCode}`))
```
asutula: Dispatch upload-specific error actions for upload errors now.

App/Sagas/ImageSharingSagas.ts
andrewxhill: pro

App/Sagas/WaitFor.ts
andrewxhill: OS it

App/Services/EventHandlers/UploadEventHandler.ts
andrewxhill: so does it also kills all other in-flight from that job if one errors?

andrewxhill:
does this get closer to limiting total number of in-flight files at any time?

App/Services/EventHandlers/UploadEventHandler.ts
asutula: Ah good question... I guess in that case, i need to retry any uploads that have errored out. Let me look into that now.

asutula:
No @andrewxhill, this is unrelated to limiting total number of in-flight files. Shouldn't be a problem with image sharing at this point. More of a thing with photo backup, which is different code.

App/Services/EventHandlers/UploadEventHandler.ts
asutula: Ok, spent some time figuring this out. Current solution is... If one of the uploads from a share fails, it allows the other uploads to continue, but marks the image share as in an 'error' state and can be retried. The retry logic examines the state of each upload involved and kicks off new uploads for any upload that hasn't started or is in an error state.

andrewxhill: that's great

PR #736, opened 2018-12-12
Handle error from mail list sign up, allow retry

### LINKED ISSUE #735: Mail list subscription not working

### LINKED ISSUE #737: Onboarding "next" button is too low in iPhone XS

asutula:
This bug where the email form submission wasn't working the first try, but would then work if you re-started the app, is a strange one. I was able to reproduce it for a while, but with a meaningless "network request failed" error. And then out of nowhere, it just started working and I can no longer reproduce it. This PR does the next best thing for now and catches the error, displays a toast with the error message, and allows the user to retry the submission. Let's see if that works.

UPDATE: Snuck in a commit to fix the arrow positioning during onboarding on iPhone X screens.

Fixes #735 
Fixes #737


asutula:
UPDATE: Snuck in a commit to fix the arrow positioning during onboarding on iPhone X screens.

andrewxhill:
perf

PR #739, opened 2018-12-13
peace of mind

andrewxhill:


App/Redux/AccountRedux.ts
```diff
@@ -38,6 +38,13 @@ const actions = {
   }),
   cafeSessionsError: createAction('CAFE_SESSIONS_ERROR', (resolve) => {
     return (error: any) => resolve({ error })
+  }),
+  refreshSDKVersionRequest: createAction('REFRESH_SDK_VERSION_REQUEST'),
+  getSDKVersionSuccess: createAction('GET_SDK_VERSION_SUCCESS', (resolve) => {
+    return (version: string) => resolve({ version })
+  }),
+  getSDKVersionError: createAction('GET_SDK_VERSION_ERROR', (resolve) => {
+    return (error: any) => resolve() // No use yet
```
asutula: TextileNodeRedux seems like the place for this.

asutula: Since it's information about the node

App/Sagas/Account/AccountSagas.ts
```diff
@@ -86,3 +87,16 @@ export function * refreshCafeSessions () {
...
   }
 }
+
+export function * getSDKVersion () {
+  while (true) {
+    try {
+      yield take(getType(AccountActions.refreshSDKVersionRequest))
+      const v: string = yield call(version)
+      yield put(AccountActions.getSDKVersionSuccess(v))
+      break
```
asutula: You don't need this break. If you keep it there, the saga will terminate for good, any future requests will be ignored. I realize there will probably not be any future requests, but if you're going to do this in the request/fulfill requests kind of pattern, might as well just support working for any request at any time.

asutula:
If you don't move it, create a ticket to do that and we'll move it later.

andrewxhill:
gtg, includes rc9

App/Sagas/NodeOnline.ts
```diff
@@ -27,5 +28,8 @@ export function * onNodeOnline () {
 
     // Check for new photos on every online event
     yield put(StorageActions.refreshLocalImagesRequest())
+
+    // update
+    yield call(getSDKVersion)
```
asutula: This was better in `NodeStarted` like you had it before

asutula: Better because if there is some problem and the node never comes online, we won't be able to see the node version while looking into the problem

PR #741, opened 2018-12-14
Update NodeOnline.ts

andrewxhill:
Gets rid of refreshMessages on each node online event


asutula:
Not really sure we want to remove that call. Seems like the one and only place we should be calling `refreshMessages`.

andrewxhill:
it was a request from @sanderpick not to do a refreshMessages onOnline

sanderpick:
you can remove since the lib does this now when it goes online. nbd though for now

PR #745, opened 2018-12-17
More migration

andrewxhill:


App/Config/ReduxPersist.ts
```diff
@@ -242,7 +242,7 @@ const persistConfig: PersistConfig = {
   key: 'primary',
   storage: AsyncStorage,
   version: 14,
-  whitelist: ['account', 'preferences', 'uploadingImages', 'processingImages', 'cameraRoll', 'storage', 'deviceLogs'],
+  whitelist: ['account', 'preferences', 'uploadingImages', 'processingImages', 'cameraRoll', 'storage', 'deviceLogs', 'migration'],
```
andrewxhill: adding to whitelist so migration data sticks around

App/Sagas/Migration.ts
```diff
@@ -32,6 +44,71 @@ export function * migrate(dispatch: Dispatch) {
...
+    const address = yield select(getAddress)
+    const username = yield select(getUsername) || ''
+    yield put(MigrationActions.announceMigration(peerId, address, username, previous.peerid))
+  }
+  // locate old peers
+  const threadItems: ThreadItem[] = yield call(getItems, THREADS_FILE_PATH)
+  const peers = [...new Set(([] as string[]).concat(...threadItems.map((thread) => thread.peers)))]
+  yield put(MigrationActions.connectToPeers(peers))
+  // run it for the first time
+  yield call(runRecurringMigrationTasks)
```
andrewxhill: run it the first time here... could also just wait for nodeOnline to run it the first time actually

App/Sagas/NodeLifecycle.ts
```diff
@@ -105,6 +105,7 @@ function * createAndStartNode(dispatch: Dispatch): any {
         yield put(PreferencesActions.migrationNeeded())
         // call the create/start sequence again
         yield call(createAndStartNode, dispatch)
+        // yield call(migrateConnections)
```
andrewxhill: needs to live somewhere

App/Sagas/NodeOnline.ts
```diff
@@ -26,5 +27,8 @@ export function * onNodeOnline () {
 
     // Check for new photos on every online event
     yield put(StorageActions.refreshLocalImagesRequest())
+
+    // Only run this after everything else in the node is running
+    yield call(runRecurringMigrationTasks)
```
andrewxhill: added here instead of lifecycle because it seems solid to just run it as the very last thing during a coming up event, since it is lowest priority

App/Containers/OnboardingScreen/index.tsx
```diff
@@ -105,11 +105,11 @@ class OnboardingScreen extends React.Component<Props, State> {
...
-              We're working fast to make Textile Photos even better for you.
-              Your old data isn't compatible with this new version of the app,
-              so for now, you'll be starting fresh. We've stored your old data
-              here on your phone, and we'll be importing it into this new
-              version of Textile Photos in a near-future update ;)
+                We're working fast to make Textile Photos even better for you.
+                Your old data isn't compatible with this new version of the app,
+                so for now, you'll be starting fresh. The migration will create
+                a local map of your private photos and your contacts so they can
+                be restored in a future release. ;)
```
asutula: Worth explaining that their contacts should begin to show up?

App/Containers/WalletPicker.js
```diff
@@ -75,9 +75,10 @@ class TextileWalletPicker extends React.PureComponent {
 const mapStateToProps = (state) => {
   const defaultData = defaultThreadData(state)
   const threadId = defaultData ? defaultData.id : undefined
-  const items = !defaultData ? [] : defaultData.photos.map((photo) => {
-    return {type: 'photo', photo, id: photo.id}
-  })
+  // const items = !defaultData ? [] : defaultData.photos.map((photo) => {
+  //   return {type: 'photo', photo, id: photo.id}
+  // })
+  const items = getSharedPhotos(state)
```
asutula: Awesome.

App/Redux/MigrationRedux.ts
```diff
@@ -79,6 +104,25 @@ export function reducer(state: MigrationState = initialState, action: MigrationA
       const { photosCount, threadsCount } = action.payload
       return { ...state, photosCount, threadsCount }
     }
+    case getType(actions.announceMigration): {
+      const announcement: PeerDetails = action.payload
+      return { ...state, announcement }
+    }
+    case getType(actions.announceSuccess): {
+      return { ...state, announcement: undefined }
```
asutula: Still need to get to the saga, but first thought here is that `announceMigration` probably doesn't need to modify the state at all. It will just trigger a saga that does a network request. I'd imagine `accounceSuccess` will be when we want to store some value that represents the fact that we've previously successfully announced the migration.

App/Sagas/Migration.ts
```diff
@@ -32,6 +44,71 @@ export function * migrate(dispatch: Dispatch) {
...
+  }
+  const peers = yield select(getNetwork)
+  for (const peer of peers) {
+    try {
+      // for each contact ask if they've migrated
+      const contact = yield call(findContact, peer)
+      if (contact) {
+        yield call(addContact, contact.peerId, contact.address, contact.username || '')
+      } else {
+        throw new Error('peer not found')
```
asutula: Couple problems here. First, don't throw an error. This is a normal and expected thing to not find the contact yet. Just let fail silently and we leave the contact in redux for future processing.

App/Sagas/Migration.ts
```diff
@@ -32,6 +44,71 @@ export function * migrate(dispatch: Dispatch) {
...
+      }
+    } finally {
+      // If no error, mark as successful
+      yield put(MigrationActions.connectionSuccess(peer))
```
asutula: Second, just put this in the `if (contact)` block. We only want to consider it a success and remove that contact from redux if we've found the contact.

App/Sagas/Migration.ts
```diff
@@ -32,6 +44,71 @@ export function * migrate(dispatch: Dispatch) {
...
+// Can be run on each node online
+export function * runRecurringMigrationTasks () {
+  const announcement = yield select(getAnnouncement)
+  if (announcement) {
+    const {peerId, address, username, previous} = announcement
+    try {
+      yield call(announceMigration, peerId, address, username, previous)
+    } finally {
+      // If no error, mark as successful
+      yield put(MigrationActions.announceSuccess())
```
asutula: Similar problem here. You're going to mark it as successful no matter what. This action should only get dispatched if we have a successful network request. You can just put this right after the `call(accounceMigration...)`. If that `call` fails it will throw an error and the `MigrationActions.announceSuccess` won't get dispatched and we'll try again in the future.

andrewxhill: ah yeah, duh

App/Redux/MigrationRedux.ts
asutula: Ok i see how you're doing this now... using the absence of `accounceMigration` as the success state. Cool.

App/Sagas/NodeLifecycle.ts
asutula: Want to enable this?

andrewxhill: done

App/Sagas/Migration.ts
```diff
@@ -32,6 +44,71 @@ export function * migrate(dispatch: Dispatch) {
...
+  // locate old peers
+  const threadItems: ThreadItem[] = yield call(getItems, THREADS_FILE_PATH)
+  const peers = [...new Set(([] as string[]).concat(...threadItems.map((thread) => thread.peers)))]
+  yield put(MigrationActions.connectToPeers(peers))
+  // run it for the first time
+  yield call(runRecurringMigrationTasks)
+}
+
+// Will error for any non-success
+export async function announceMigration(peerId: string, address: string, username: string, previousId: string) {
```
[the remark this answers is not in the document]
andrewxhill: fixed

App/Containers/OnboardingScreen/index.tsx
andrewxhill: yeah, let's revisit before we release to see how far migration gets

App/Sagas/NodeOnline.ts
asutula: Ok last thing here. The way this is set up, if you happen to get another `online` event while a previous `runRecurringMigrationTasks` saga is running, you're going to get repeated calls to the various apis for the same data. If that isn't harmful at all, cool. If you think it's a problem, this can be fixed by dispatching some `requestrunRecurringMigrationTasks` action and having a saga that listens for it using `take`, blocking while it yields the `runRecurringMigrationTasks` saga, then loops around to `take` again. You can see this in sagas i've written that start with `while (true) { yield take(...) ... }`

andrewxhill:
cool. it works without the username included, and once added it'll continue to work. right now it's setup as an optional. so going to merge and update framework whenever.

PR #749, opened 2018-12-19
Adds dev script to just launch the both textile and rn daemons

andrewxhill:


asutula:
Should it edit the `.env` with peer id and gateway url?

andrewxhill:
not sure... didn't want to get too tricky, but we could

PR #750, opened 2018-12-19
quickfix

andrewxhill:


asutula:
Ah yea good catch. Hate those indexes, pretty lame way to do it, but it was fast.

PR #752, opened 2018-12-21
Photo migration as an option

andrewxhill:


asutula:
Cool dude, good to get started on this. One thing I want to clear up about the general approach so we can get on the same page...

My original work on this was designed around the idea that the photo migration (or any long running part of the migration) would be executed though a blocking UX where we, all in a blocking modal dialog, explain that the migration will happen and that they need to leave the app running in the foreground, execute the migration with some sort of progress reporting, and the dump the user in the the regular app UI.

The reason for this was to keep the migration as simple as possible and not have to deal with background downloads (the current migration saga doesn't use background downloads) and questions about "what if the user leaves the app and then...?".

Seems like you're work here makes the photo migration a thing that can run while the user can use the app as normal, and therefore, is more likely to background or quit the app. 

I think that is a great UX to aim for, but I'm just concerned that we'll end up with a lot more work to do to get it right and deal with all the "what if"s.

Thoughts?

andrewxhill:
that's not true, you may have missed my original chats so maybe if you have a second go back in slack to a few days ago. i was hoping you could jump in here and put your intended waiting screen up after the user clicks the notification.

asutula:
Ah the thing that pops up is a "notification" in the image feed? Was trying to figure out what that was and made the wrong conclusion that it was to actually UI that triggered AND displayed the progress of the migration.

asutula:
Sorry, i mean notification feed

asutula:
Is this something we're wanting to include in the initial migration release or in a later release? 

If the answer is in the initial migration release, we could more easily just present the option to migrate photos during the migration/onbaording flow. 

If the answer is in a later release, the notification solution is a good one.

andrewxhill:
nah, yeah, it's just a notification in the feed to let you know it needs to happen. from there your ui + progress etc could launch. right now it goes right into your saga, but ideally you'll do your thing there

andrewxhill:
i didn't know how far we would get so was eliminating any dependencies and just moving forward.

asutula:
Ok, I'll try to get the actual migration code done asap and we'll decide the best place to plug it into the UI

andrewxhill:
another benefit of leaving it in the notifications is you can warn the user that you'll need some bandwidth for a minute and let them dismiss until later (if they are on bad or costly data).

asutula:
Yea good point

asutula:
Made good progress on migration today. Everything is basically working. Need to get this guy fixed before we can release: https://github.com/textileio/textile-go/issues/434

Remaining tasks for tomorrow are to implement the cleanup logic that cleans up all the files and artifacts created by migration in the case that the migration is complete or the user opts out of migration.

asutula:
Ok I think this is ready for review @andrewxhill

asutula:
Cool, just going to get the waitlist signup in today then I'll merge.

Issue #762, opened 2018-12-28
Improve query for unique shared items


asutula:
Sander said something about using the top level target. Look into this, sounds more simple.

andrewxhill:
Right, so this should pretty simple now. 

### A little background

Back when we first were having these issues, we didn't have the pre-hash so we didn't know where to keep track of an index of prior added photos. Now, the node is doing that for us, we should be okay.

Back when we didn't have it though, I was polling and adding photos more conservatively, ignore updates etc so as to avoid more of the duplicates. We can loosen that up now once we have a check on uniqueness instead.

### Thinking

For now, I think that you can get that from here `processingPhoto.preparedFiles.dir.files[0].checksum` also maybe here, `processingPhoto.preparedFiles.dir.files[0].source`, I'm not certain what source does and didn't see it in the docs. See question here, https://github.com/textileio/docs/issues/84

It will be tricky for a client to then decide what to do with that information.

1. They could scan against all their in-memory hashes. -1
2. We could maybe add variable in the original payload that says, `exists: bool`, but that isn't necessarily enough information to decide what to do;
3. What if it's a photo that someone else had shared in a Thread and now you've added to your camera roll independently? In this case, you'd probably still want it backed up in your account camera roll thread, (e.g. if they un-pin it, you don't).

andrewxhill:
Ah, apologies, you might not be talking about Camera Roll adds here, but instead actual photos shared. We can move that comment, I just know you're starting to think through the above so wanted to get that background down.

Issue #761, opened 2018-12-28
Support sharing from phone camera roll and different thread schemas


asutula:
We now have shared potentially coming from three places:

1. Phone camera roll
2. Camera roll thread
3. Shared threads

Will need to refactor the model and logic to deal with that.

PR #763, opened 2018-12-28
Update to rc17

asutula:
Looks to be working. Some changes are in the "good enough for now" category. Created issues to track those potential improvements.

### DIFF
--- a/App/Redux/PhotoViewingSelectors.ts
+++ b/App/Redux/PhotoViewingSelectors.ts
@@ -104,7 +104,7 @@ export function getSharedPhotos (state: RootState): ReadonlyArray<SharedPhoto> {
       .filter((photo) => photo.author_id === selfId)
       .map((photo): SharedPhoto => {
         const files = photo.files.length ? photo.files[0] : undefined
-        const original = files && files.links ? files.links.exif.checksum : photo.block
+        const original = files && files.links ? files.links.thumb.checksum : photo.block
         return { type: 'photo', photo, id: photo.block, original }
       })
     )
--- a/App/SB/components/ThreadDetailCard/index.tsx
+++ b/App/SB/components/ThreadDetailCard/index.tsx
@@ -182,7 +182,7 @@ const mapStateToProps = (state: RootState, ownProps: OwnProps): StateProps => {
   // Unsquares the images by maintaining the aspect ratio no matter device size
   const imageWidth = WIDTH
   const links = photo.files[0].links
-  const meta = links ? links['large'].meta : undefined
+  const meta = links ? links['thumb'].meta : undefined
   const width = meta ? meta['width'] as number : undefined
   const height = meta ? meta['height'] as number : undefined
   let imageHeight = imageWidth
--- a/App/Sagas/ImageSharingSagas.ts
+++ b/App/Sagas/ImageSharingSagas.ts
@@ -118,9 +118,9 @@ export function * prepareImage (uuid: string) {
     }
     const { sharedImage, destinationThreadId } = processingImage
     const preparedFiles: IMobilePreparedFiles = yield call(prepare, sharedImage, destinationThreadId)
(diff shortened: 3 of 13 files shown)


App/Redux/PhotoViewingSelectors.ts
```diff
@@ -104,7 +104,7 @@ export function getSharedPhotos (state: RootState): ReadonlyArray<SharedPhoto> {
       .filter((photo) => photo.author_id === selfId)
       .map((photo): SharedPhoto => {
         const files = photo.files.length ? photo.files[0] : undefined
-        const original = files && files.links ? files.links.exif.checksum : photo.block
+        const original = files && files.links ? files.links.thumb.checksum : photo.block
```
asutula: @sanderpick this is the code that's pulling the unique list of shared photos (regardless of thread or comments, likes, etc)

asutula: Just made this minimal change for now, created #762 to improve

App/SB/components/ThreadDetailCard/index.tsx
```diff
@@ -182,7 +182,7 @@ const mapStateToProps = (state: RootState, ownProps: OwnProps): StateProps => {
   // Unsquares the images by maintaining the aspect ratio no matter device size
   const imageWidth = WIDTH
   const links = photo.files[0].links
-  const meta = links ? links['large'].meta : undefined
+  const meta = links ? links['thumb'].meta : undefined
```
asutula: This is a bit of a hack now... switched all these metadata queries to `thumb` since shared and private threads both have a `thumb` link and for now so these screens can be used to display images from either type of thread. Improvements tracked in #764

App/Sagas/ImageSharingSagas.ts
```diff
@@ -118,9 +118,9 @@ export function * prepareImage (uuid: string) {
     }
     const { sharedImage, destinationThreadId } = processingImage
     const preparedFiles: IMobilePreparedFiles = yield call(prepare, sharedImage, destinationThreadId)
-    if (sharedImage.isAvatar && preparedFiles.dir && preparedFiles.dir.files && preparedFiles.dir.files['large'].hash) {
+    if (sharedImage.isAvatar && preparedFiles.dir && preparedFiles.dir.files && preparedFiles.dir.files['raw'] && preparedFiles.dir.files['raw'].hash) {
       // TODO: This doesn't seem right in here, but ok
-      const hash = preparedFiles.dir.files['large'].hash as string
+      const hash = preparedFiles.dir.files['raw'].hash as string
```
asutula: Camera roll thread no longer has a `large` link, so switch to `raw` here

App/Sagas/NodeLifecycle.ts
```diff
@@ -93,7 +93,7 @@ function * createAndStartNode(dispatch: Dispatch): any {
     const cameraRollThreadKey = Config.RN_TEXTILE_CAMERA_ROLL_THREAD_KEY
     const cameraRollThread = threadsResult.find((thread) => thread.key === cameraRollThreadKey)
     if (!cameraRollThread) {
-      yield call(addThread, cameraRollThreadKey, cameraRollThreadName)
+      yield call(addThread, cameraRollThreadKey, cameraRollThreadName, false)
```
asutula: Camera roll thread being created with `shared === false`

App/Sagas/PhotoViewingSagas.ts
```diff
@@ -67,7 +67,7 @@ export function * addThread (action: ActionType<typeof PhotoViewingActions.addTh
   const { name } = action.payload
   try {
     const key = `textile_photos-shared-${uuid()}`
-    yield call(add, key, name)
+    yield call(add, key, name, true)
```
asutula: All other threads created in the app with `shared === true`

asutula:
For now, we can support sharing from one shared thread to another, but not from the camera roll thread to a shared thread. That will require some better data modeling and calling `prepareFIles` any time the schema the photo is being shared from is different than the schema the photo is being shared to. Tracked in #761

PR #771, opened 2019-01-02
"Threads" to "Groups" in UI

andrewxhill:


App/Containers/ThreadsList.js
```diff
@@ -107,10 +107,10 @@ class ThreadsList extends React.PureComponent {
           style={onboardingStyles.emptyStateImage}
           source={require('../Images/v2/thread-empty-state.png')} />
         <Text style={onboardingStyles.emptyStateText}>
-          This is your shared photo stream,
+          This is your private update stream,
```
asutula: Word 'private' is a little confusing here maybe?

andrewxhill: was just trying to make a smidge obvious it isn't a public feed a la twit/insta

PR #801, opened 2019-01-06
Clicking comment now goes to Photo Screen

andrewxhill:
Saw this again and just wanted to nip it.

Makes more sense, when you click a comment you want to see what the commented as well as what they commented on. 

Based it of @asutula's suggestion, just link into the PhotoScreen but don't limit the comment count.

### DIFF
--- a/App/Containers/PhotoScreen.tsx
+++ b/App/Containers/PhotoScreen.tsx
@@ -41,7 +41,7 @@ class PhotoScreen extends React.Component<StateProps & NavigationScreenProps<{}>
           item={{type: 'photo', photo: this.props.photo, threadId: this.props.threadId, threadName: this.props.threadName}}
           onComment={this.onComment}
           onLikes={this.onLikes}
-          recentCommentsCount={5}
+          recentCommentsCount={-1}
           maxLinesPerComment={5}
         />
       }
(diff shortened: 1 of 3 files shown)


App/Containers/PhotoScreen.tsx
```diff
@@ -41,7 +41,7 @@ class PhotoScreen extends React.Component<StateProps & NavigationScreenProps<{}>
           item={{type: 'photo', photo: this.props.photo, threadId: this.props.threadId, threadName: this.props.threadName}}
           onComment={this.onComment}
           onLikes={this.onLikes}
-          recentCommentsCount={5}
+          recentCommentsCount={-1}
```
asutula: Just using an optional for `ThreadDetailCard.props.recentCommentsCount` and then not setting here would be nice.

App/SB/components/ThreadDetailCard/index.tsx
```diff
@@ -84,6 +84,7 @@ class ThreadDetailCard extends React.PureComponent<OwnProps & StateProps & Dispa
     } = this.props
 
     const { photo } = item
+    const recentCommentsCount = this.props.recentCommentsCount === -1 ? photo.comments.length : this.props.recentCommentsCount
```
asutula: Then this would just be a check for if it is defined or not.

App/Containers/PhotoScreen.tsx
andrewxhill: updated

PR #813, opened 2019-01-10
Node logs screen

asutula:
Get's pretty brutal on memory usage to read in the whole log (can be many MB), so reading in the last 300KB for now. Provides refresh and copy to clipboard. Let's see how that feels.
screen shot 2019-01-10 at 15 29 38

### DIFF
--- a/App/Sagas/NodeLifecycle.ts
+++ b/App/Sagas/NodeLifecycle.ts
@@ -143,7 +143,6 @@ function * createAndStartNode(dispatch: Dispatch): any {
         yield put(AccountActions.setRecoveryPhrase(recoveryPhrase))
         yield put(TextileNodeActions.derivingAccount())
         const walletAccount: WalletAccount = yield call(walletAccountAt, recoveryPhrase, 0)
-        const logToDisk = !__DEV__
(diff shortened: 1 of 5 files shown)


App/Sagas/NodeLifecycle.ts
```diff
@@ -143,9 +143,8 @@ function * createAndStartNode(dispatch: Dispatch): any {
         yield put(AccountActions.setRecoveryPhrase(recoveryPhrase))
         yield put(TextileNodeActions.derivingAccount())
         const walletAccount: WalletAccount = yield call(walletAccountAt, recoveryPhrase, 0)
-        const logToDisk = !__DEV__
         yield put(TextileNodeActions.initializingRepo())
-        yield call(initRepo, walletAccount.seed, REPO_PATH, logToDisk)
+        yield call(initRepo, walletAccount.seed, REPO_PATH, true)
```
asutula: Seems like we always want to write logs to disk because it's useful

asutula:
If you don't want to copy all the contents, you can also long press to get the OS 'select/copy' UX.

PR #820, opened 2019-01-11
allow un to wrap on wallet screen when long

### LINKED ISSUE #702: Long usernames cut off on wallet view.
Screen capture from user reporting this issue: 
9a85edc9-3d84-477f-9399-27760bd49834

andrewxhill:
fixes #702 

worth testing on android @carsonfarmer because it relies on window dimensions which sometimes aren't perfect over there

but probably don't need to delay for that i don't think


carsonfarmer:
Cool, can test this shortly.

PR #824, opened 2019-01-11
Block multi comment

andrewxhill:


andrewxhill:
this guy next

App/Containers/Comments.tsx
```diff
@@ -57,6 +67,24 @@ class Comments extends Component<Props> {
...
+      // the comment was flushed so we can type a new one and submit again
+      this.setState({submitting: false})
+    } else if (this.props.commentError === true && this.state.submitting === true) {
+      // there was an error, allow a retry
+      this.setState({submitting: false})
+    }
+  }
+
+  onSubmit = () => {
+    return () => {
```
asutula: You don't have to do these nested functions. Just eliminate the nested one, and in the `onSubmit` binding below, make it like: `onSubmit={this.onSubmit}`

andrewxhill: fixed

PR #829, opened 2019-01-15
rc25 + contacts migration updates

asutula:


App/NativeModules/Textile/Api.ts
```diff
@@ -32,12 +32,9 @@ export async function acceptThreadInviteViaNotification(id_: string): Promise<st
...
 
-export async function addContact(id_: string, address: string, username: string): Promise<void> {
-  try {
-    await TextileNode.addContact(id_, address, username)
-  } catch (error) {
-    // nothing for now
-  }
+export async function addContact(contact: ContactInfo): Promise<void> {
+  const contactJsonString = JSON.stringify(contact)
+  await TextileNode.addContact(contactJsonString)
```
asutula: Updated to pass the string-ified `ContactInfo` to the go layer to add a new contact

App/NativeModules/Textile/Model.ts
```diff
@@ -106,10 +106,12 @@ export interface CafeSession {
...
-  readonly username: string
-  readonly inboxes: ReadonlyArray<Cafe>
-  readonly added: string
-  readonly thread_ids: ReadonlyArray<string>
+  readonly username?: string
+  readonly avatar?: string
+  readonly inboxes?: ReadonlyArray<Cafe>
+  readonly created: string
+  readonly updated: string
+  readonly thread_ids?: ReadonlyArray<string>
```
asutula: This `ContactInfo` model needed updating. Using it for both `addContact` and the return value for `contact()` and `contacts()` since the optional `thread_ids` is the only thing that doesn't exist on the go layer's `Contact` model.

App/Redux/ContactsRedux.ts
```diff
@@ -30,7 +30,7 @@ export function reducer (state: ContactsState = initialState, action: ContactsAc
 
 export const ContactsSelectors = {
   isKnown: (state: RootState, id: string) => state.contacts.contacts.some((p) => p.id === id),
-  byThreadId: (state: RootState, id: string) => state.contacts.contacts.filter((contact) => contact.thread_ids.indexOf(id) > -1)
+  byThreadId: (state: RootState, id: string) => state.contacts.contacts.filter((contact) => (contact.thread_ids || []).indexOf(id) > -1)
```
asutula: `thread_ids` is an optional array in go, so made the corresponding model change in TS... had some ripple effects like this and a few other places in the app's code.

App/Sagas/Migration.ts
```diff
@@ -204,12 +204,11 @@ export function * announcePeer() {
   if (previousItems.length) {
     const previous = previousItems[0]
     const peerId = yield select(getPeerId)
-    const address = yield select(getAddress)
+    const contactInfo: ContactInfo = yield call(contact, peerId)
```
asutula: Here we call the go layer's `contact()` method with our own peer id, so it returns a `ContactInfo` for the local peer. That is what we announce and send to the server.

App/Sagas/Migration.ts
```diff
@@ -255,12 +253,13 @@ export function * runRecurringMigrationTasks () {
...
 
 // Will error for any non-success
-async function announceId(currentPeerId: string, previousPeerId: string, currentAddress: string, previousUsername?: string) {
+async function announcePeerDetails(peerDetails: PeerDetails) {
   const headers = {'Content-Type': 'application/json'}
-  const payload: { [key: string]: string } = { peerId: currentPeerId, previousId: previousPeerId, address: currentAddress }
-  const username = previousUsername ? previousUsername.trim() : ''
-  if (username.length > 0) {
-    payload['username'] = username
+  const contactInfoJsonString = JSON.stringify(peerDetails.currentContactInfo)
```
asutula: Sending a JSON string of the `ContactInfo` so we can store it as a string in the database

App/Sagas/Migration.ts
```diff
@@ -270,10 +269,15 @@ async function announceId(currentPeerId: string, previousPeerId: string, current
...
 
 // will error response doesn't include the peer
-async function findContact(peerId: string): Promise<{peerId: string, previousId: string, address: string, username?: string}> {
+async function findContact(peerId: string): Promise<ContactInfo | undefined> {
   const response = await fetch(`${Config.RN_PEER_SWAP}?peerId=${peerId}`, { method: 'GET' })
-  const responseJson: [{peerId: string, previousId: string, address: string, username?: string}] = await response.json()
-  return responseJson[0]
+  const peerDetailsArray: ReadonlyArray<{ previousPeerId: string, previousUsername?: string, currentContactInfo: string }> = await response.json()
+  if (peerDetailsArray.length > 0) {
+    const contactInfo = JSON.parse(peerDetailsArray[0].currentContactInfo) as ContactInfo
```
asutula: Parsing the persisted string from the server back into a JSON object here

PR #834, opened 2019-01-15
Local avatar when node started

andrewxhill:
…il online

### DIFF
--- a/App/Components/Avatar.tsx
+++ b/App/Components/Avatar.tsx
@@ -14,7 +14,7 @@ interface OwnProps {
 
 interface StateProps {
   localPeerId?: string
-  online: boolean
+  nodeStarted: boolean
   peerColor: string
   localAvatar?: string
 }
@@ -61,7 +61,7 @@ class Avatar extends React.Component<Props, State> {
       height = h
     }
 
-    const { peerId, localPeerId, online, localAvatar } = this.props
+    const { peerId, localPeerId, nodeStarted, localAvatar } = this.props
 
     const isLocalUser = (!peerId && localPeerId) || peerId === localPeerId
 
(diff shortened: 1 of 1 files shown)


App/Components/Avatar.tsx
```diff
@@ -133,7 +133,7 @@ const mapStateToProps = (state: RootState, ownProps: OwnProps): StateProps => {
 
   return {
     localPeerId: state.account.peerId.value,
-    online: state.textileNode.online,
+    nodeStarted: state.textileNode.nodeState.state === 'started',
```
sanderpick: Wouldn't this be false for all other states? nodeState can be...
```
'nonexistent' = 'nonexistent',
  'creating' = 'creating',
  'created' = 'created', // Node has been created, on it's way to starting
  'starting' = 'starting',
  'started' = 'started',
  'stopping' = 'stopping',
  'stopped' = 'stopped', // Node has been explicitly stopped, different than created
  'creatingWallet' = 'creatingWallet',
  'derivingAccount' = 'derivingAccount',
  'initializingRepo' = 'initializingRepo'
```

andrewxhill: yes. can we be fetching the image data when the node is in other states?

sanderpick: it needs to be at least started, so you'd need something like the online flag that toggles to true, right?

sanderpick: you could do the started state OR online

andrewxhill: sorry, i haven't been following the node state convos closely enough. eli5. this currently would be `false` but if the node reaches `started` it would toggle to `true`. if later it changed to anything else, `stopped` or `stopping` whatever, it would go back to `false` and avoid the ui trying to make requests by hash. which other states would you want to be `true`?

sanderpick: oh, apologies... i assumed there was an online node state as well... but i guess it's just the flag. ignore!

andrewxhill:
was thinking about that too, but seems backburner for now

Issue #831, opened 2019-01-15
Update Avatar flow is confusing/broken


sanderpick:
The flow of choosing a photo and the state of the save button is confusing / broken. Here's what I see:

- Tap avatar in wallet header
- Tap Choose Photo
- Tap Choose from library (haven't tried from Camera, but prob same)
- Immediately, the big avatar on the change screen goes white, but the Save button _is not blocked_
- After some time, the new avatar appears (maybe we can show a spinner in the blank white circle?)
- Whether or not i wait for the new avatar to appear, when I hit Save, the button just changes back to “Choose Photo”
- If I hit the back button, it appears as if the avatar was updated
- After a hard kill, the new avatar is visible (so, it's working, just a flow issue)

Ideally, we can keep the user on this screen until the avatar is fully updated. A little spinner would help.

carsonfarmer:
Outlook is less grim on-device. Resolution of new avatar is slow, but the actually flow itself seems to work. But as @sander points out, it 'feels' like you've just be dumped back on the previous step. Maybe we could disable the buttons, or move you to save but have it disabled until the image loads?

andrewxhill:
that's the way it's setup to work. it fires a goBack but somehow it's broken. fixed to navigate to to wallet screen explicitly in my PR

PR #838, opened 2019-01-15
goBack seemed to fail here

andrewxhill:
changing to explicitly go back to wallet made it work again

### DIFF
--- a/App/Containers/Wallet.js
+++ b/App/Containers/Wallet.js
@@ -164,7 +164,7 @@ class Wallet extends React.PureComponent {
       <View style={style.container}>
         <WalletHeader
           changeAvatar={() => {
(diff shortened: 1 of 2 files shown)


App/Containers/Wallet.js
```diff
@@ -164,7 +164,7 @@ class Wallet extends React.PureComponent {
       <View style={style.container}>
         <WalletHeader
           changeAvatar={() => {
-            this.props.navigation.navigate('ChangeAvatar', { onSuccess: () => this.props.navigation.goBack() })
+            this.props.navigation.navigate('ChangeAvatar', { onSuccess: () => this.props.navigation.navigate('Wallet') })
```
andrewxhill: This fixes #831. In this case goBack didn't result in a navigation.

PR #842, opened 2019-01-16
Android: avatar border radius fix

### LINKED ISSUE #837: Avatar component does not handle border-radius
At least when it's using `TextileImage`. Note, this is in the emulator.

sanderpick:
- Adds a `View` wrapper to `TextileImage` the includes the border radius, `overflow: hidden` is also needed
- Updates android gradle plugin to latest (3.3.0), which requires gradle 4.10.1

fixes #837

### DIFF
--- a/App/Components/Avatar.tsx
+++ b/App/Components/Avatar.tsx
@@ -74,14 +74,16 @@ class Avatar extends React.Component<Props, State> {
     if (isLocalUser && nodeStarted && localAvatar && widthNumber) {
       // Render TextileImage
       return (
-        <TextileImage
-          style={{ ...(this.props.style || {}), width, height, borderRadius: this.state.borderRadius }}
-          target={localAvatar}
-          index={0}
-          forMinWidth={widthNumber}
-          resizeMode={'cover'}
-          onLayout={this.onImageLayout}
-        />
+        <View style={{ width, height, borderRadius: this.state.borderRadius, overflow: 'hidden' }}>
+          <TextileImage
+            style={{ ...(this.props.style || {}), width, height, borderRadius: this.state.borderRadius }}
+            target={localAvatar}
+            index={0}
+            forMinWidth={widthNumber}
+            resizeMode={'cover'}
+            onLayout={this.onImageLayout}
+          />
+        </View>
       )
     } else if (uri) {
       const uniqueUserColor = !this.state.showIcon ? {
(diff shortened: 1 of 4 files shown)


sanderpick:
Seeing these DIFF bundle requests with latest gradle when doing JS reloads... definitely helps the sanity level.

screenshot 2019-01-16 06 02 43

App/Components/Avatar.tsx
```diff
@@ -74,14 +74,16 @@ class Avatar extends React.Component<Props, State> {
...
+        <View style={{ width, height, borderRadius: this.state.borderRadius, overflow: 'hidden' }}>
+          <TextileImage
+            style={{ ...(this.props.style || {}), width, height, borderRadius: this.state.borderRadius }}
+            target={localAvatar}
+            index={0}
+            forMinWidth={widthNumber}
+            resizeMode={'cover'}
+            onLayout={this.onImageLayout}
+          />
+        </View>
```
asutula: Did you test if the same border radius problem exists in the other cases of this `if` statement? I'm guessing it does... When I was looking into this yesterday, the problem seemed to be that the `onLayout={this.onImageLayout}` (which results in the border radius being set) wasn't being triggered on Android. The same technique is used in those other `if` cases.

android/app/build.gradle
```diff
@@ -124,8 +124,8 @@ android {
         applicationIdSuffix ".dev"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 0
-        versionName "0"
+        versionCode 1
```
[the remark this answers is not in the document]
asutula: Should be fine, it's changed by the build process. This default value is only uses when you install to and emulator or device locally.

App/Components/Avatar.tsx
sanderpick: I didn't explicitly... but good point. Though I did see a URI based avatar and a blank Icon one (before URI resolved) from my iOS beta install, and they were both looked fine. We could wrap the whole if else block in a `View` just to be sure. Re: layout event... the border radius is coming through so it must be fired at some point. I logged it in render at one point... was zero for a couple renders, then good values.

asutula: Cool, yea there's a good chance that when i saw it not firing, it was just because of all the problems with AS and emulators and devices. I really can't explain how broken everything Android seemed yesterday. It was very confusing.

sanderpick: K, cool. You ok w/ not doing the full View wrap unless we see that issue? I pretty sure the other cases aren't an issue... probably why we hadn't seen this before all over the place.

asutula: yea sounds good to me

PR #849, opened 2019-01-17
Node logs fix

asutula:
Fixes log reading offset bug, clear out in memory data when loading more log data just to avoid memory pressure crashes. I think it was probably the offset bug causing our crash, though.


asutula:
Also set the log level to DEBUG for local dev builds, but ERROR for beta and release builds

asutula:
Oh actually, would it be better to have the Fabric Beta builds be logging at DEBUG? I think we can actually do that with our .env based setup.

asutula:
Ok now using .env `RN_RELEASE_TYPE` so local dev builds and Fabric beta builds use DEBUG, production builds ERROR.

sanderpick:
Awesome dude. After we merge/release your mobile api for dynamic logs, we could also switch to DEBUG when you long press the logo. Could be great for remote debugging with a real user.

PR #853, opened 2019-01-19
PubSub Contacts

### LINKED ISSUE #851: Padding around comments is no longer rendering
Old:
screenshot 2019-01-17 10 34 50

New:
screenshot 2019-01-17 10 35 18

sanderpick:
- Updates models and API for https://github.com/textileio/textile-go/pull/459
- Update avatar usage to expect the target hash from components
- Cleans-up Avatar component
- fixes #851

### DIFF
--- a/App/Components/Avatar.tsx
+++ b/App/Components/Avatar.tsx
@@ -66,11 +66,12 @@ class Avatar extends React.Component<Props, State> {
     const { borderRadius: radius } = this.state
 
     if (icon || !started || !target) {
+      const heightNumber = typeof height === 'number' ? height as number : defaultSize
       return (
         <Icon
           style={{ ...(this.props.style || {}), width, height, borderRadius: radius }}
           name={icon || 'question-circle'}
-          size={height as number}
+          size={heightNumber}
           color={s.COLOR_GREY_LIGHT}
           onLayout={this.onImageLayout}
         />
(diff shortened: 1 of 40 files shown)


sanderpick:
Still some work and testing to do here... I'm sure I missed some UI updates.

sanderpick:
Fixes https://github.com/textileio/textile-mobile/issues/851

App/Components/Avatar.tsx
```diff
@@ -49,96 +51,83 @@ class Avatar extends React.Component<Props, State> {
...
-        <View style={{ width, height, borderRadius: this.state.borderRadius, overflow: 'hidden' }}>
+        <View style={{ ...(this.props.style || {}), width, height, borderRadius: radius, overflow: 'hidden' }}>
           <TextileImage
-            style={{ ...(this.props.style || {}), width, height, borderRadius: this.state.borderRadius }}
-            target={localAvatar}
+            style={{ ...(this.props.style || {}), width, height, borderRadius: radius }}
+            target={target}
             index={0}
-            forMinWidth={widthNumber}
+            forMinWidth={width as number}
```
asutula: Going to have some problems here if width was set as a string, like `'50%'`, for example. That's why it used to use `widthNumber` and wouldn't render the `TextileImage` unless width is a number.

asutula:
Going to be some nasty conflicts to resolve between this one and @andrewxhill's currently open PR (he moved many of the filed edited in this PR to other repos).

App/Components/Avatar.tsx
sanderpick: Ok, cool. I'll add a check back in

sanderpick:
Hmm, right. Well, sounds like the SDK PR needs a little re-thinking in terms of what components are broken out... would be easier to c/p the contents from master rather than propagating these changes to all their new homes. As in, I think it will be easier if this goes in first.

sanderpick:
Ok @asutula - added some checks for width/height as a number when those are needed. Just falling back to `defaultSize`.

PR #875, opened 2019-01-23
Adds textile image to package

andrewxhill:
Must not have committed these changes and now looks missing on master. Redid it here.


asutula:
Looks like you just had it as a different name `import io.textile.rnmobile.RNTextilePackage;`. I'm running master now and seems to work fine.

asutula:
Actually, I take that back... Haven't run it yet, but it does look like that import is there under another name, right?

asutula:
Ok sorry for the play by play as I understand more of what's going on here...

So, this adds `TextileImage` in a new package (as opposed to the node, which is provided by `RNTextileImage`). It's not clear to me where the code for that package comes from. Is there an npm module that provides it? 

Either way, seems like the code for `TextileImage` should be bundled up in the same repo as the node, etc, and that it would just be provided by the same package, `RNTextilePackage`.

andrewxhill:
I had mentioned this the other day, I didn't pull TextileImage out of textile-mobile yet because I hit a block with creating UIViews in modules and wiring them up correctly. So in both iOS and Android TextileImage remained in this textile-mobile repo. 

By removing TextileNode and the packager that wired it up in Android, it left the TextileImageView/Manager classes dangling and uncallable. This rewires them into the main app so they are callable from the RN code.

asutula:
Just curious, what was the snag you hit adding these to the sdk repo/package?

andrewxhill:
I didn't seem to have them registering correctly in iOS so `requireNativeComponent` would just error and say not available

PR #878, opened 2019-01-24
Convert to Drawer Nav

asutula:
The goal here is to convert the current content of the app to a drawer navigator, getting ready for the new UX and screens we've been planning.

The goal here is _not_ to create amazing new screens with the best UX. Looks like:

screen shot 2019-01-24 at 12 41 13

screen shot 2019-01-24 at 12 41 09

screen shot 2019-01-24 at 12 41 18

Now that the settings menu open from the drawer, I converted it to open as a modal over the open drawer.


asutula:
I should also explain that this removes the Timeline screen we had in favor of a Groups screen which was created from the groups list in the old wallet screen. When you tap on a group in the Groups screen, it opens the timeline for that single group. This is a step in the direction we are going with the new UX plan. 

Similarly, I converted the contacts grid from the old wallet screen into a full contacts screen that is accessible from the drawer. We'll be implementing a new contacts screen next.

PR #881, opened 2019-01-28
Mock SDK lifecycle management methods

andrewxhill:


App/Redux/AccountRedux.ts
```diff
@@ -2,6 +2,9 @@ import { createAction, ActionType, getType } from 'typesafe-actions'
 import { ContactInfo, CafeSession } from '@textile/react-native-sdk'
```
andrewxhill: Account stuff I considered as all part of SDK logic

App/Redux/MockBridge.ts
```diff
@@ -0,0 +1,39 @@
+import { createAction, getType, ActionType } from 'typesafe-actions'
```
andrewxhill: MockBridge is just creating methods that were otherwise missing or needed to call to pull app logic out of sdk functions.

App/Sagas/NodeLifecycle.ts
```diff
@@ -2,42 +2,32 @@ import { delay, Task } from 'redux-saga'
...
   registerCafe,
   cafeSessions,
   CafeSession
  } from '@textile/react-native-sdk'
-import { logNewEvent } from './DeviceLogs'
-import { announcePeer } from './Migration'
 
 export const REPO_PATH = `${RNFS.DocumentDirectoryPath}/textile-go`
-const MIGRATION_NEEDED_ERROR = 'repo needs migration'
+// const MIGRATION_NEEDED_ERROR = 'repo needs migration'
```
andrewxhill: Needed to ignore migration for a minute. Too interwoven between app / sdk

App/Sagas/NodeLifecycle.ts
```diff
@@ -236,31 +222,22 @@ function * backgroundTaskRace () {
 
 function * stopNodeAfterDelay (ms: number) {
   try {
-    if (yield select(PreferencesSelectors.verboseUi)) {
-      yield call(displayNotification, 'Running the node for 20 sec. in the background')
-    }
+
+    yield put(MockBridgeActions.stopNodeAfterDelayStarting())
```
andrewxhill: Good example of an event well need to create from the SDK that is just mocked here by mockbridge

App/Sagas/NodeStarted.ts
```diff
@@ -2,14 +2,14 @@ import { call, put, take } from 'redux-saga/effects'
...
 import AccountActions from '../Redux/AccountRedux'
 import ContactsActions from '../Redux/ContactsRedux'
-import PreferencesActions from '../Redux/PreferencesRedux'
 import TextileNodeActions from '../Redux/TextileNodeRedux'
 import PhotoViewingActions from '../Redux/PhotoViewingRedux'
 import { getSDKVersion } from './NodeLifecycle'
 
 export function * onNodeStarted () {
-  while (yield take([getType(TextileNodeActions.startNodeSuccess), getType(PreferencesActions.onboardedSuccess)])) {
+  while (yield take([getType(TextileNodeActions.startNodeSuccess), getType(AccountActions.initSuccess)])) {
```
andrewxhill: Don't want to rely on the ui's onboarding success, here just rely on if the account was initialized yet or not. I think the sdk is going to want a different 'stop' / 'start' state than the one returned by the node, that simply lets the app turn off the node long term even though it's init'd. will cross that bridge later.

andrewxhill:
I think next step here is actually a review prior to merge. Here are the full steps to complete the work,

1. Review
2. Merge
3. Live test in team Beta
4. Migrate all `App/SDK` to the react-native-sdk
5. Update react-native-sdk version
6. Release

PR #920, opened 2019-02-08
Isolating missed online event

andrewxhill:


App/Config/ReduxPersist.ts
```diff
@@ -300,7 +300,7 @@ const migrations: MigrationManifest = {
 const persistConfig: PersistConfig = {
   key: 'primary',
   storage: AsyncStorage,
-  version: 16,
+  version: 17,
```
andrewxhill: that was kinda dumb

App/Sagas/Account/AccountSagas.ts
```diff
@@ -53,9 +54,9 @@ export function * setUsername () {
   while (true) {
     try {
       const action: ActionType<typeof AccountActions.setUsernameRequest> = yield take(getType(AccountActions.setUsernameRequest))
-      const online: boolean = yield call(Textile.nodeOnline)
-      if (!online) {
-        yield take(getType(TextileEventsActions.nodeOnline))
+      const nodeState = yield select(TextileEventsSelectors.nodeState)
```
andrewxhill: doesn't matter if node is online, eventually this could be moved to sdk

App/Sagas/NotificationsSagas.ts
```diff
@@ -156,10 +157,10 @@ export function * reviewThreadInvite (action: ActionType<typeof NotificationsAct
 
 export function * waitUntilOnline(ms: number) {
```
andrewxhill: should drop this function

App/Sagas/TextileEventsSagas.ts
```diff
@@ -3,19 +3,15 @@ import { ActionType, getType } from 'typesafe-actions'
...
 import { RootAction } from '../Redux/Types'
-import Config from 'react-native-config'
 import Textile, {
-  ContactInfo,
-  ThreadInfo
+  ContactInfo
  } from '@textile/react-native-sdk'
 import { logNewEvent } from './DeviceLogs'
-import { pendingInvitesTask } from './ThreadsSagas'
+import { pendingInvitesTask, cameraRollThreadCreateTask } from './ThreadsSagas'
```
andrewxhill: moved that logic to it's own function, felt like we crammed it into a specific loop but it shoudl be on it's own, it's not even used in current app ever

PR #929, opened 2019-02-10
Expose a single event subscription method

andrewxhill:
Instead of having to make the client mix native and js event subscriptions, I'm just managing within the SDK now and exposing a single interface to add/remove subscriptions. Also helps to create more unique event keys that wont mess with client events while still not making the client have to subscribe to long/ugly keys.


andrewxhill:
ah, also as some of those js events get migrated to native modules, it will remain backward compatible by not making the client care what type they are in RN world

asutula:
Going to close this, replicated the work on my sdk update branch

PR #937, opened 2019-02-21
Splits TextileImage out into a reusable package

andrewxhill:


andrewxhill:
this one is gtg now too

PR #939, opened 2019-02-26
Update SDK

asutula:
This brings in many more protobuf types, gives us STACK mode for group feeds so we can start developing new UI for them, and provides streaming contact search results.

Deletes some old/unused files as well.


asutula:
Ok, big one, good to go!

PR #940, opened 2019-02-26
Feed UI crash fix

asutula:
You get the crash if you try to render 0 length text in a `Text` component. If you effectively do this:
```
<Text>{''}</Text>
```
the UI will crash.


asutula:
Need to be careful about this now that empty strings can come from the Go layer in many places

PR #947, opened 2019-02-28
Simplify onboarding code, remove possible bugs

asutula:
The approach here was to clean up and simplify:

- Remove band aid fixes that haven't worked
- Remove array splicing based on redux state, always show referral code screen, pre-populated with a referral code when applicable
- Remove a couple things that might have been causing bugs

### DIFF
--- a/App/Components/ReferralCode.tsx
+++ b/App/Components/ReferralCode.tsx
@@ -59,7 +59,8 @@ const HIT_SLOP: Insets = {
 }
 
 interface Props {
-  referralCode: string
+  targetReferralCode: string
+  referralCode?: string
   onSuccess?: () => void
 }
 
@@ -77,15 +78,16 @@ export default class ReferralCode extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props)
     this.state = {
-      valid: false,
-      showWaitlistSignup: false
+      valid: props.referralCode ? props.referralCode.trim().toLowerCase() === this.props.targetReferralCode.trim().toLowerCase() : false,
+      showWaitlistSignup: false,
+      referralCode: props.referralCode
     }
   }
 
   updateText = (text: string) => {
     this.setState({
       referralCode: text,
-      valid: text.trim().toLowerCase() === this.props.referralCode.trim().toLowerCase()
+      valid: text.trim().toLowerCase() === this.props.targetReferralCode.trim().toLowerCase()
     })
   }
 
--- a/App/Containers/OnboardingScreen/index.tsx
+++ b/App/Containers/OnboardingScreen/index.tsx
@@ -29,7 +29,7 @@ const ARROW_FORWARD: ViewStyle = {
 }
 
 interface StateProps {
-  skipReferralCode: boolean
+  referralCode?: string
 }
 
 interface DispatchProps {
@@ -39,7 +39,6 @@ interface DispatchProps {
 type Props = StateProps & DispatchProps & NavigationScreenProps
 
 interface State {
-  blockNext: boolean
   showArrow: boolean
   currentPage: number
 }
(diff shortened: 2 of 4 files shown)


App/Components/ReferralCode.tsx
```diff
@@ -59,7 +59,8 @@ const HIT_SLOP: Insets = {
 }
 
 interface Props {
-  referralCode: string
+  targetReferralCode: string
+  referralCode?: string
```
asutula: The referral code input component now optionally takes a prop that is a referral code (set in the case the user is onboarding from opening the app from an invite link)

App/Containers/OnboardingScreen/index.tsx
```diff
@@ -29,7 +29,7 @@ const ARROW_FORWARD: ViewStyle = {
 }
 
 interface StateProps {
-  skipReferralCode: boolean
+  referralCode?: string
```
asutula: No more skipping referral code component

App/Containers/OnboardingScreen/index.tsx
```diff
@@ -39,7 +39,6 @@ interface DispatchProps {
 type Props = StateProps & DispatchProps & NavigationScreenProps
 
 interface State {
-  blockNext: boolean
```
asutula: Removing this fix that didn't seem to help

App/Containers/OnboardingScreen/index.tsx
```diff
@@ -72,19 +70,20 @@ class OnboardingScreen extends React.Component<Props, State> {
...
+        showArrow: this.showArrowForIndex(this.state.currentPage + 1)
+      })
       this.pages.scrollToPage(this.state.currentPage + 1)
-      setTimeout(() => {
-        this.setState({
-          blockNext: false,
-          showArrow: this.showArrowForIndex(this.state.currentPage + 1),
-          currentPage: this.state.currentPage + 1
-        })
-      }, 350)
```
asutula: `nextPage` now simply tells the pager component to scroll to the next page and sets the arrow visibility based on what the next page will be.

App/Containers/OnboardingScreen/index.tsx
```diff
@@ -72,19 +70,20 @@ class OnboardingScreen extends React.Component<Props, State> {
...
     }
   }
 
+  onScrollEnd = () => {
+    this.setState({
+      currentPage: this.state.currentPage + 1
+    })
+  }
+
```
asutula: Using the `onScrollEnd` callback from the pager component to update `state.currentPage` once we've actually arrived on that page.

App/Containers/OnboardingScreen/index.tsx
```diff
@@ -149,10 +149,6 @@ class OnboardingScreen extends React.Component<Props, State> {
         />
       )
     ]
-    // remove referral code if needed.
-    if (this.props.skipReferralCode) {
-      pages.splice(3, 1)
-    }
```
asutula: No more dynamic splicing of the page data. It it always the same now.

carsonfarmer: 👍

App/Containers/OnboardingScreen/index.tsx
```diff
@@ -166,8 +162,8 @@ class OnboardingScreen extends React.Component<Props, State> {
             style={[CONTAINER]}
             containerStyle={{ marginBottom: spacing._016 }}
             indicatorColor={color.accent2_2}
-            startPage={this.state.currentPage}
```
asutula: This is actually the thing I'm most suspicious could have caused problems. This `render` method gets called over and over and over again. `startPage` is, in theory, something that should be static, but we were setting it to `state.currentPage` which is changing as the user progresses through the onboarding. You would think that the pager component should deal with this correctly, but there's a good change it doesn't and may be causing our bug.

asutula: So, omitting this prop, it defaults to `0`, and we never pass in any other value.

PR #957, opened 2019-03-08
Big SDK update

asutula:
This is ready to go except for it's blocked by this potential `go-textile` bug where `Files` annotations are always empty in the `feed` api: https://github.com/textileio/go-textile/issues/603. This causes the group feed to not show likes or comments on photos.


asutula:
Ok finally, g2g!

PR #965, opened 2019-03-18
Updates to rc 1.1.0-rc42

andrewxhill:


andrewxhill:
putting this in so we can test it out

PR #969, opened 2019-03-19
Update sdk/go-textile

andrewxhill:


andrewxhill:
should be gtg

asutula:
Pretty weird there is no yarn.lock update in here

PR #983, opened 2019-03-25
Fix inverted flatlist

andrewxhill:
If you want to replicate these changes in your branch we can close it. But i've just tested and these updates make the Android scrolling on the flatlist work again.

issue was described here https://github.com/facebook/react-native/issues/19434, it causes a nasty bounce effect while scrolling up an inverted list


asutula:
Looks like this involves just update react native? If so, let's leave this and wait for my branch where I already updated react native.

andrewxhill:
yep, that's it. we can leave open until yours lands

asutula:
@andrewxhill I just made the remaining RN update changes from https://github.com/react-native-community/rn-diff-purge/compare/version/0.57.2...version/0.57.7 and pushed them here. G2G.

PR #984, opened 2019-03-25
removes old migration field

andrewxhill:


asutula:
Nice. Did you already remove all the peer swap code as well?

andrewxhill:
there was nothing in the code that referenced that field at least

asutula:
Yea, i just meant code that was triggering and calling the endpoints

andrewxhill:
nah, nothing else in this one except cleaning the .env

PR #985, opened 2019-03-25
Only show one avatar if a block of messages in a row

andrewxhill:
AndroidScreen

IMG_57545B2F902F-1


App/screens/group/group.tsx
```diff
@@ -137,7 +138,24 @@ class Group extends Component<Props, State> {
...
+  sameUserAgain = (user: pb.IUser, previous: Item): boolean => {
+    if (!previous || !previous.type) {
+      return false
+    }
+    switch (previous.type) {
+      case 'message': {
+        if (user.name.length <= 0 || previous.data.user.name.length <= 0) {
+          return false
+        }
+        return user.name === previous.data.user.name
```
asutula: Don't think we want to use `name` since it's not unique to a user. `address` would be it.

andrewxhill: ah yeah, good call

PR #986, opened 2019-03-25
Fixes Type.Message click from Notifications

andrewxhill:


asutula:
Looks like you didn't push any changes and the change in here is unrelated to the fix

Issue #962, opened 2019-03-15
Some local photo copies aren't being cleaned up


andrewxhill:
If you download the app folder you can see them in

`AppData/Documents/images`

it's an odd mix of a dozen or so for me

andrewxhill:
There is another folder, I think this must have been fixed

`AppData/documents/go-textile/images`

with a different dozen images, these are all oldish though

Issue #887, opened 2019-02-01
Updating (or setting) Avatar on Android leads to a duplicate original photo


andrewxhill:
**Describe the bug**
When you set your avatar on an android device, the photo you select will be duplicated and stored again to disk. 

**To Reproduce**
1. Go to Textile Photos
2. Select update Avatar
3. Choose a photo already stored on your device
4. Confirm
5. Now select to update again and when you browse your photos you'll notice a 2nd copy

PR #998, opened 2019-03-28
Ensure app doesn't create duplicate photo copies on android

andrewxhill:
fixes https://github.com/textileio/textile-mobile/issues/887

fixes https://github.com/textileio/textile-mobile/issues/962


andrewxhill:
k, tested on ios, works fine

PR #1015, opened 2019-04-05
Update TextileImage; add Avatar caching

andrewxhill:
Uses the image cache option to try and force it, but that's an incomplete solution. so here will also use the local ipfs node to grab avatars, meaning the app should look a lot better when on spotty internet or no connection


andrewxhill:
scratch the ipfs image part, just undid that. round about way of getting there but realized the avatar component was too aggressively refreshing content. when app goes into background, we don't want to trigger a refresh actually, so fixed that

PR #1021, opened 2019-04-06
adds back a required permission of the camera picker for android

andrewxhill:


android/app/src/main/AndroidManifest.xml
```diff
@@ -28,7 +27,8 @@
         android:name=".MainActivity"
         android:label="@string/app_name"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
-        android:windowSoftInputMode="adjustResize">
+        android:windowSoftInputMode="adjustResize"
+        android:launchMode="singleInstance">
```
asutula: This is so weird, but ok cool. Not sure why this would not be the default. Related to this... I've noticed when i tap a deep link from Slack, the Textile activity acutally starts up "inside" the Slack process. It's very strange. Anyone else?

android/app/src/main/AndroidManifest.xml
```diff
@@ -9,7 +9,6 @@
     <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
     <!-- Remove some unused RN default permissions -->
     <uses-permission tools:node="remove" android:name="android.permission.READ_PHONE_STATE" />
-    <uses-permission tools:node="remove" android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
```
asutula: Just commenting for now only because this PR is titled "adding back permission" and all I see here is removing a permission. Just want to be sure this is the intention.

android/app/src/main/AndroidManifest.xml
andrewxhill: I've fixed that for Android (that is what this fix does exactly).
For iOS I don't know the solution yet.

android/app/src/main/AndroidManifest.xml
andrewxhill: "tools:node='remove'" was removing the permission, removing the line adds it back as it is a default added by RN.

andrewxhill: oh, i do know the solution and it's schema based links. https will embed, textile-dev: will break out. that is why i added that accept page in http so that if a user lands there it will textile: schema link them and they can break out.

andrewxhill: ^ for ios. for android it was easier.

android/app/src/main/AndroidManifest.xml
asutula: I didn't think it' a problem in iOS. No?

android/app/src/main/AndroidManifest.xml
asutula: Ah got it. Thx.

android/app/src/main/AndroidManifest.xml
andrewxhill: doh, sorry, added new comment on wrong place

PR #1023, opened 2019-04-08
Cleans up image loading for Android

andrewxhill:
…pe images are explicitly set; reduces the shouldUpdate calls on textileimage types


andrewxhill:
@asutula I'm going to bang this guy out to android users as the app is pretty slow to use without it

PR #1024, opened 2019-04-08
Update to RN sdk 1.1.4 + Android textile deps from maven

asutula:
Does a few things:

1. Updated to RN sdk 1.1.4 which pulls in `android-textile` (native sdk), `mobile` (the generated aar), and `pb` (the generated pb classes) via gradle/maven.
2. Updates the android app project to use multidex because that is necessary now (https://developer.android.com/studio/build/multidex)
3. Removes the yarn dependency on `go-mobile` and it's associated project configuration to resolve the .aar because we no longer need that... We're getting that dependency from gradle/maven now via the RN sdk.


android/app/build.gradle
```diff
@@ -173,26 +174,26 @@ android {
 }
 
 dependencies {
-    compile project(':@textile_react-native-sdk')
+    implementation 'com.android.support:multidex:1.0.3'
+    implementation project(':@textile_react-native-sdk')
```
asutula: `compile` is deprecated, so updating all those to `implementation`.

android/build.gradle
```diff
@@ -25,6 +25,9 @@ allprojects {
         mavenLocal()
         google()
         jcenter()
+        maven {
+            url "https://dl.bintray.com/textile/maven"
+        }
```
asutula: This repo is necessary in order to resolve our `mobile` maven package (the aar).

android/settings.gradle
```diff
@@ -27,4 +27,3 @@ include ':react-native-screens'
 project(':react-native-screens').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-screens/android')
 include ':react-native-contacts'
 project(':react-native-contacts').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-contacts/android')
-include ':mobile'
```
asutula: No more of this wiring in of the aar in the client project necessary

PR #1027, opened 2019-04-09
updates tslint rules; upgrades type sources

andrewxhill:


package.json
```diff
@@ -41,10 +41,11 @@
     "redux-persist": "^5.10.0",
     "redux-saga": "^0.16.1",
     "stacktrace-js": "^2.0.0",
-    "typesafe-actions": "^2.0.4",
+    "typesafe-actions": "3.4.0",
```
andrewxhill: Okay, updating some of our source types.

Also updating the base tslint to a shared repo to reuse across the projects.

tslint.json
```diff
@@ -38,27 +38,11 @@
     "max-line-length": [true, 200]
   },
   "rules": {
+    "member-access": false,
```
andrewxhill: reduces the customization beyond the shared rule-set

App/Components/GroupCard/index.tsx
```diff
@@ -63,8 +65,8 @@ const GroupCard = (props: ScreenProps) => {
       </View>
       <View style={styles.groupRightColumn}>
         <View style={styles.avatarContainer}>
-          { members.slice(0, 8).map((mem: pb.IContact, i: number) => {
-            const imageStyle = cardImageStyle(members.length, i)
+          { memberArray.slice(0, 8).map((mem: pb.IContact, i: number) => {
```
andrewxhill: This was a hidden bug I guess... will open a ticket to investigate the source type being labeled incorrectly as an array

package.json
```diff
@@ -54,7 +55,7 @@
     "@types/react-native-fs": "^2.8.2",
     "@types/react-native-push-notification": "^3.0.3",
     "@types/react-native-qrcode": "^0.2.2",
-    "@types/react-redux": "^6.0.9",
+    "@types/react-redux": "7.0.6",
```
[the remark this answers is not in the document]
asutula: Ok, did the research... Looks like you can go ahead and update `react-redux` to 7.0.6 and it should be compatible with our current versions of React and Redux. Go for it.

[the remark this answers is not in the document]
andrewxhill: just bumped react-redux to 6.x. It's react requirement is 'React 16.4 or later' which we already satisfy. no issues that i see

PR #1029, opened 2019-04-10
fixing crashing errors

andrewxhill:


App/Components/PhotoGrid.tsx
```diff
@@ -78,6 +78,8 @@ class PhotoGrid extends React.Component<ScreenProps & DispatchProps & Navigation
     const { item } = row
     switch (item.type) {
       case 'photo':
+        const files = item.photo.files
+        const fileIndex = files && files.length > 0 && files[0].index ? files[0].index : 0
```
andrewxhill: not sure how this could happen, but it matches the pattern of some crashes we were seeing

App/Containers/Groups.tsx
```diff
@@ -242,7 +243,7 @@ class Groups extends React.PureComponent<Props, State> {
 }
 
 const mapStateToProps = (state: RootState): StateProps => {
-  const ownAddress = state.account.address.value
+  const ownAddress = getAddress(state)
```
andrewxhill: moving these to selectors, they might also require a non-null check in the chain... shouldn't but our crashes indicate something null happening that shouldn't be

PR #1031, opened 2019-04-10
avoids cases of dynamic empty strings in text elements

andrewxhill:


App/SB/components/FeedItem/index.tsx
```diff
@@ -48,6 +48,9 @@ const FeedItem = (props: FeedItemProps) => {
     </View>
   )
 
+  const feedText = payload.feed && payload.feed !== '' ? payload.feed : undefined
```
asutula: I'm not sure what happens if you render `undefined` in a `Text` component. Might get the same crash?

andrewxhill: the ticket i shared says, no

asutula: Think one of us is reading that wrong. To me, it's talking about that in the the condition `{this.state.error && ` will only work as intended when `error` is `undefined` and not `''`. They are avoiding every rendering an empty string by putting some static text in there with a possibly `undefined` `this.state.error`:
```
<Text>
    Error message: {this.state.error}
  </Text>
```

asutula: So there is always a string with length in the `Text` component

App/SB/components/FeedItem/index.tsx
```diff
@@ -60,11 +63,10 @@ const FeedItem = (props: FeedItemProps) => {
         </View>
       </View>
       <View style={styles.textContainer}>
-        <Text style={styles.text}>{payload.feed}</Text>
+        <Text style={styles.text}>{feedText}</Text>
```
asutula: So in this case, we only want to render this `Text` element if `feedText` is  defined and has a length > 0

PR #1033, opened 2019-04-10
Spring cleaning

andrewxhill:


andrewxhill:
okay, that's gtg. not sure much to be done with a review though

PR #1038, opened 2019-04-13
alphabetize contact list

andrewxhill:


asutula:
Should we have a better default ordering the go layer?

andrewxhill:
okay, ticket added

PR #1045, opened 2019-04-17
Update to RN SDK 2

asutula:
Still wrapping up some last things, but wanted to get this out there so anyone interested could see what kind of updates are needed in an app that migrates to the new RN SDK.

Biggest changes seemed to be:

- subscribing to events
- no more `pb` prefix on model classes
- need to register cafes (no logic provided by SDK) (still wip, coming asap)


App/Components/GroupCard/index.tsx
```diff
@@ -3,16 +3,16 @@ import { View, Text, TouchableOpacity } from 'react-native'
...
 import styles, { cardImageStyle, ICON_WIDTH, ROW_COLUMN } from './statics/styles'
 import { color } from '../../styles'
 
 interface ScreenProps {
   id: string
   name: string
-  members: pb.IContact[]
-  thumb?: pb.IFiles
+  members: IContact[]
+  thumb?: IFiles
```
asutula: Ok the removal of the `pb` namespace created tons of noise in this PR. I'll try to leave comments where anything actually interesting changed.

App/Redux/TextileEventsRedux.ts
```diff
@@ -1,42 +1,34 @@
 import { createAction, getType, ActionType } from 'typesafe-actions'
-import { NodeState } from '@textile/react-native-sdk'
 import { RootState } from './Types'
 
+export enum NodeState {
+  stopped = 'stopped',
+  started = 'started',
+  online = 'online'
+}
```
asutula: No more `NodeState` model provided by the sdk, just event subscriptions for node started, stopped, etc, so we model the state here to store in redux and use in the app.

App/Redux/TextileEventsRedux.ts
```diff
@@ -1,42 +1,34 @@
...
-    (resolve) => (type: string, message: string) => resolve({ type, message })
-  ),
-  startNodeFinished: createAction(
-    '@bridge/startNodeFinished',
-    (resolve) => () => resolve()
-  ),
   stopNodeAfterDelayStarting: createAction(
     '@bridge/stopNodeAfterDelayStarting',
-    (resolve) => () => resolve()
+    (resolve) => (delay: number) => resolve({ delay })
```
asutula: The callback for running the node in the background provides the amount of time it will run for, which is dynamically calculated on iOS.

App/Redux/TextileEventsRedux.ts
```diff
@@ -56,30 +48,35 @@ interface TextileEventsState {
...
+      return { ...state, nodeState: { ...state.nodeState, error } }
     default:
       return state
   }
 }
 
 export const TextileEventsSelectors = {
   nodeState: (state: RootState) => state.textile.nodeState,
-  online: (state: RootState) => state.textile.online
+  started: (state: RootState) => state.textile.nodeState.state === NodeState.started || state.textile.nodeState.state === NodeState.online,
```
asutula: I'm modeling `online` as another `NodeState`, so when the node is "started" it can be in either a `started` or `online` state... This selector makes it easy to ask if the node is "started".

App/Sagas/Account/AccountSagas.ts
```diff
@@ -47,7 +48,7 @@ export function * refreshPeerId() {
   while (true) {
     try {
       yield take(getType(AccountActions.refreshPeerIdRequest))
-      const peerIdResult = yield call(API.ipfs.peerId)
+      const peerIdResult = yield call(Textile.ipfs.peerId)
```
asutula: No more `API` export from the sdk. The whole API just available from the default export (called `Textile` here). This matches the ios and android sdk structure exactly.

App/Sagas/TextileEventsSagas.ts
```diff
@@ -88,25 +86,6 @@ export function * ignoreFileRequest() {
...
-      if (yield select(PreferencesSelectors.showNodeStateNotification)) {
-        yield call(displayNotification, 'App State Change: ' + action.payload.newState)
-      }
-
-      yield call(logNewEvent, 'State Change', action.payload.newState)
-    } catch (error) {
-      yield call(logNewEvent, 'appStateChange', error.message, true)
-    }
-  }
-}
```
asutula: "App state" is no longer exposes by the sdk (it's used internal to the native sdks, but that is an implementation detail that could change in the future). It's irrelevant to an end user app, and it's something the developer can access themselves using existing platform apis.

App/Services/EventHandlers/TextileNodeEventHandler.ts
```diff
@@ -10,112 +9,108 @@ import { contactsActions } from '../../features/contacts'
...
-      if (type_url === '/Comment' ||
-        type_url === '/Like' ||
-        type_url === '/Files' ||
-        type_url === '/Ignore' ||
-        type_url === '/Join') {
-        this.store.dispatch(PhotoViewingActions.refreshThreadRequest(update.thread))
-      }
-
-      if (type_url === '/Join' ||
+    this.subscriptions.push(
```
asutula: New events subscription api used in here.

App/Services/EventHandlers/TextileNodeEventHandler.ts
```diff
@@ -10,112 +9,108 @@ import { contactsActions } from '../../features/contacts'
...
+      })
+    )
+    this.subscriptions.push(
+      Textile.events.addThreadRemovedListener((threadId) => {
+        this.store.dispatch(PhotoViewingActions.threadRemoved(threadId))
+      })
+    )
+    this.subscriptions.push(
+      Textile.events.addNotificationReceivedListener((notification) => {
+        this.store.dispatch(NotificationActions.newNotificationRequest(toTypedNotification(notification)))
```
asutula: The new event subscription API provides callbacks with typed data (`notification` here is a `INotification` type)

App/Services/EventHandlers/TextileNodeEventHandler.ts
```diff
@@ -10,112 +9,108 @@ import { contactsActions } from '../../features/contacts'
...
+      Textile.events.addCanceledPendingNodeStopListener(() => {
+        this.store.dispatch(TextileEventsActions.stopNodeAfterDelayCancelled())
+      })
+    )
   }
 
   tearDown() {
-    this.events.removeAllListeners()
+    for (const subscription of this.subscriptions) {
+      subscription.cancel()
```
asutula: Event subscriptions return an instance of something called `EventSubscription` on which you can cancel the subscription later.

andrewxhill:
i'll give this a try in notes in a few too. only thing that really off in the last update was that my nodeStarted listener (new kind) never seemed to fire even though it was clearly running.

asutula:
@andrewxhill you have your code pushed anywhere I can see it? One problem could be, you need to hang on to the `EventSubscription` that is returned from `addListener` I think.

asutula:
Meh, I'm like 60% about that statement actually.

PR #1048, opened 2019-04-24
More enhancements to the join retry flow

andrewxhill:
Couple things I saw playing more with it.

- Remembers threadName on a retry request
- If progress was dismissed and then user tries to accept the invite again, it will undismiss the progress so they have some feedback


andrewxhill:
okay, this guy is gtg. slipped in ui for feedback on notification based invite accept.

flow is, click notification accept -> sends user to the Groups screen so they can see the joining feedback -> opens the group if it's successful

PR #1049, opened 2019-04-24
Basic app based thread filtering

andrewxhill:


andrewxhill:
- slipped in ui feedback for internal invite joining status now too

PR #1057, opened 2019-05-12
Avatar/profile fixes/updates

asutula:
For now, managing updates to the avatar using `Textile.files.prepare/add`. In the future, the go layer may provide a more specific `setAvatar` method.

Avatar/profile management had turned into quite a rat's nest and was almost impossible to understand. I ended up consolidating all avatar/profile related code into an "account" feature in `features/account`. Much easier to follow now, but it caused to to have to touch lots of files, making this PR kinda big.

Also realized in the process and after talking with @sanderpick that the old `setAvatar` method never did anything that led to the avatar data being remotely pinned. This PR corrects that, the data is pinned to the cafes.

Also used the opportunity to start laying the groundwork for upcoming "file sync" work, but it not really used for anything at this point.


asutula:
Snuck a fix in to display any node start/stop error along with the node state in verbose UI mode.

asutula:
Any takers?

PR #1065, opened 2019-05-15
Adds Remove Contacts Functionality

### LINKED ISSUE #1051: Remove a contact
Per user request, you should be able to delete a contact from your "address book". The API support this: `removeContact`

undercase:
Fixes #1051. Adds the ability to remove a contact from the `Contact` detail view/modal. To this end, I made a few other notable changes:

- The `Contact` view now accepts an `IContact` object as a parameter, instead of `avatar`, `username`, and `address`.

- There is now a button in the `Contact` view that will let the user add a contact (if it is not already in the user's contacts) or remove a contact (if it is).

- The Contact profile layout is now vertical, not horizontal, so that more functionality can be added without disrupting the layout.

### DIFF
--- a/App/Containers/Contact.tsx
+++ b/App/Containers/Contact.tsx
@@ -40,7 +40,7 @@ interface DispatchProps {
   addContact: (contact: IContact) => void
 }
 
-type Props = StateProps & NavigationScreenProps<NavProps>
+type Props = StateProps & DispatchProps & NavigationScreenProps<NavProps>
 
 class ContactModal extends React.Component<Props> {
 
--- a/App/Containers/Contact.tsx
+++ b/App/Containers/Contact.tsx
@@ -36,8 +36,8 @@ interface StateProps {
 }
 
 interface DispatchProps {
-  removeContact: (address: string) => void
-  addContact: (contact: IContact) => void
+  removeContact: () => void
+  addContact: () => void
 }
 
 type Props = StateProps & DispatchProps & NavigationScreenProps<NavProps>
@@ -101,11 +101,11 @@ class ContactModal extends React.Component<Props> {
   }
 
   onRemove = () => {
-    this.props.removeContact(this.props.navigation.getParam('contact').address)
+    this.props.removeContact()
   }
 
   onAdd = () => {
-    this.props.addContact(this.props.navigation.getParam('contact'))
+    this.props.addContact()
   }
 }
 
--- a/App/Containers/Contact.tsx
+++ b/App/Containers/Contact.tsx
@@ -65,13 +65,10 @@ class ContactModal extends React.Component<Props> {
 
   render() {
     const avatar = this.props.navigation.getParam('contact').avatar
-    let buttonText: string
-    if (this.props.isContact) {
-      buttonText = this.props.removing ? 'Removing' : 'Remove'
-    } else {
-      buttonText = this.props.adding ? 'Adding' : 'Add'
-    }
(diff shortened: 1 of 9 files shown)


App/Containers/Contact.tsx
```diff
@@ -67,29 +80,59 @@ class ContactModal extends React.Component<Props> {
...
+    isContact,
+    removing,
+    adding
+  }
+}
+
+const mapDispatchToProps = (dispatch: Dispatch<RootAction>): DispatchProps => {
+  return {
+    removeContact: (address: string) => dispatch(contactsActions.removeContactRequest(address)),
+    addContact: (contact: IContact) => dispatch(contactsActions.addContactRequest(contact))
```
asutula: One small, but nice improvement is using the optional `ownProps` parameter in `mapDispatchToProps`. Then you don't have to have `address: string` and `contact: IContact` as parameters to these prop functions... Inside the functions, you can just pull `address` and `contact` from `ownProps`.

App/features/contacts/actions.ts
```diff
@@ -50,6 +50,22 @@ export const clearAddContact = createAction('@contacts/CLEAR_ADD_CONTACT_ERROR',
...
+  return (address: string) => resolve({ address })
+})
+
+export const removeContactSuccess = createAction('@contacts/REMOVE_CONTACT_SUCCESS', (resolve) => {
+  return (address: string) => resolve ({ address })
+})
+
+export const removeContactError = createAction('@contacts/REMOVE_CONTACT_ERROR', (resolve) => {
+  return (address: string, error: any) => resolve({ address, error })
+})
```
asutula: It's good to use `typesafe-actions`' `createAsyncAction` for these request, success, error setups. It's cleaner. You can find examples in other places in our codebase.

App/features/contacts/actions.ts
```diff
@@ -50,6 +50,22 @@ export const clearAddContact = createAction('@contacts/CLEAR_ADD_CONTACT_ERROR',
...
+
+export const clearRemoveContactError = createAction('@contacts/CLEAR_REMOVE_CONTACT_ERROR', (resolve) => {
```
asutula: Let's not use the `@` prefix on those strings, not sure how/why I started that.

undercase:
Fixed the issues you brought up. Thanks 👍.

App/Containers/Contact.tsx
```diff
@@ -1,36 +1,46 @@
...
   threadThumbs: ReadonlyArray<ThreadThumbs>
+  isContact: boolean
+  removing: boolean
+  adding: boolean
 }
 
-type Props = StateProps & NavigationScreenProps<NavProps>
+interface DispatchProps {
+  removeContact: (address: string) => void
+  addContact: (contact: IContact) => void
```
asutula: So you still need to remove the arguments from these.

App/Containers/Contact.tsx
```diff
@@ -67,29 +80,61 @@ class ContactModal extends React.Component<Props> {
...
     )
   }
+
+  onRemove = () => {
+    this.props.removeContact(this.props.navigation.getParam('contact').address)
+  }
+
+  onAdd = () => {
+    this.props.addContact(this.props.navigation.getParam('contact'))
+  }
```
asutula: And same in the calls inside of these

undercase: Oof idk how I missed this. Fixing it right now.

PR #1068, opened 2019-05-16
update camera roll fix

asutula:
Signed-off-by: Aaron Sutula <hi@asutula.com>


asutula:
Pulls in a version of camera roll that declares android deps in a better way

PR #1077, opened 2019-05-21
Only query camera roll if pref says to

asutula:


App/features/photos/sagas.ts
```diff
@@ -14,9 +14,8 @@ import { ProcessingPhoto } from './models'
 
 function * toggleStorage(action: ActionType<typeof PreferencesActions.toggleStorageRequest>) {
   const { name } = action.payload
-  const storageOption = yield select(PreferencesSelectors.storage, name)
-  const storageStatus = !storageOption ? false : storageOption.status
```
asutula: The storage options are always defined (they're all defined in the default state of the reducer), so simplified this a bit.

asutula:
Pushed the sdk update into here as well @andrewxhill

PR #1084, opened 2019-05-22
Add Support for Renaming Threads

undercase:
From a group's action menu, choose "Rename Group." This opens a modal. The modal closes automatically once the new name has been submitted successfully.


.gitignore
```diff
@@ -49,6 +49,7 @@ local.properties
 node_modules/
 npm-debug.log
 yarn-error.log
+package-lock.json
```
asutula: You should really use yarn! We all do, so there is no lockfile for npm checked in, only `yarn.lock`. So maybe I should say, you have to use yarn ;)

App/Containers/RenameGroupModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+const mapDispatchToProps = (dispatch: Dispatch<RootAction>, ownProps: NavigationScreenProps<NavProps>) : DispatchProps => {
+  const threadId = ownProps.threadId
+  return {
+    rename: (newName: string) => { dispatch(groupActions.renameGroup.renameGroup.request({ threadId, name: newName })) }
+  }
+}
+
+export const RenameGroupComponent = connect(mapStateToProps, mapDispatchToProps)(RenameGroupScreen)
+
+export default class RenameGroupModal extends React.Component<ScreenProps & ModalProps> {
```
asutula: Not really seeing the need for a while class to wrap the modal. You can just put this in the `render` of `RenameGroupScreen`

App/Containers/RenameGroupModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+  rename: (newName: string) => void
+}
+
+interface ScreenProps {
+  cancel: () => void
+  complete: () => void
+}
+
+interface ModalProps {
+  isVisible: boolean
```
asutula: This can just be part of `ScreenProps` if you do that.

App/Containers/Styles/RenameGroup.js
```diff
@@ -0,0 +1,43 @@
+import { StyleSheet } from 'react-native'
+
+import Colors from '../../Themes/Colors'
+import { color } from '../../styles'
+
+export default StyleSheet.create({
```
asutula: Let's just define all styles as `const`s within the file that uses them. And use explicit typescript types when defining them. There are plenty of example of this in the code base, but let me know if that doesn't make sense.

App/Containers/Styles/RenameGroup.js
```diff
@@ -0,0 +1,43 @@
...
+  container: {
+    flex: 1,
+    backgroundColor: Colors.modalColor,
+    flexDirection: 'column',
+    justifyContent: 'center',
+    alignItems: 'stretch',
+    padding: 20
+  },
+  inputStyle: {
+    height: 60,
```
asutula: For all the sizes, margins, paddings, etc, please use the const vales defined in the `styles` module. Similar to how you dealt with colors here.

asutula: Same with font family.

App/Containers/Styles/RenameGroup.js
```diff
@@ -0,0 +1,43 @@
+import { StyleSheet } from 'react-native'
+
+import Colors from '../../Themes/Colors'
+import { color } from '../../styles'
+
+export default StyleSheet.create({
+  container: {
+    flex: 1,
+    backgroundColor: Colors.modalColor,
```
asutula: This is some old color module i'm trying to phase out. Use the new one like below.

App/features/group/add-message/actions.ts
```diff
@@ -1,7 +1,20 @@
 import { createAsyncAction } from 'typesafe-actions'
 
+export interface addMessagePayload {
```
asutula: Typically type/interface name are Capitalized. Maybe we should find a lint rule for that.

App/features/group/reducer.ts
```diff
@@ -4,19 +4,22 @@ import { ActionType } from 'typesafe-actions'
...
 }
 
 export type GroupAction = ActionType<typeof actions>
 
 export default combineReducers<GroupState>({
   feed: feedReducer,
   addMessage: addMessageReducer,
-  addPhoto: addPhotoReducer
+  addPhoto: addPhotoReducer,
+  renameGroup: renameGroupReducer
```
asutula: Good work groking the nested feature setup ;)

App/features/group/rename-group/reducer.ts
```diff
@@ -0,0 +1,39 @@
...
+        ...state,
+        [threadId]: {
+          error: errorMessage
+        }
+      }
+    }
+    default:
+      return state
+  }
+}
```
asutula: I know they don't already exist everywhere, but let's try to add a test for any new reducer at least. There are some examples in the code base already, but let me know if you have questions.

App/features/group/rename-group/sagas.ts
```diff
@@ -0,0 +1,26 @@
...
+    yield call(Textile.threads.rename, threadId, name)
+    yield put(renameGroup.success({ threadId }))
+    yield put(PhotoViewingActions.refreshThreadRequest(threadId))
+  } catch (error) {
+    yield put(renameGroup.failure({ threadId, error }))
+  }
+}
+
+export function *watchForRenameGroupRequests() {
+  yield takeEvery(getType(renameGroup.request), handleRenameGroupRequest)
```
asutula: Been meaning to get rid of the pattern of having this "watch" function. Just put `takeEvery(getType(renameGroup.request), handleRenameGroupRequest)` directly into the `yield all([])` below.

.gitignore
undercase: Okay, I'll use yarn!

App/features/group/reducer.ts
undercase: Thanks 😄.

undercase:
@asutula Fixed everything you mentioned above and added my first tests. Taking this out of draft mode now that textile-go is fixed.

App/features/group/rename-group/sagas.ts
```diff
@@ -0,0 +1,22 @@
...
+import { call, put, takeEvery, all } from 'redux-saga/effects'
+import Textile from '@textile/react-native-sdk'
+import { renameGroup } from './actions'
+import PhotoViewingActions from '../../../Redux/PhotoViewingRedux'
+
+export function *handleRenameGroupRequest(action: ActionType<typeof renameGroup.request>) {
+  const { threadId, name } = action.payload
+  try {
+    yield call(Textile.threads.rename, threadId, name)
+    yield put(PhotoViewingActions.updateThreadName(threadId, name))
```
asutula: We have this situation now where the list of threads is in a different place in the redux tree than the group data. You can see the consequence of that here where you have to dispatch this random action for another reducer. I've been trying to think of a nice way to consolidate these two sources of information. Will be good to figure out soon.

undercase: I think it'd be ideal to just have a single threads reducer that stores all thread/group related data-plus refactoring the two branches of the redux tree together would give us the opportunity to normalize some stuff like using `createAsyncAction`.

PR #1087, opened 2019-05-22
Make Links Clickable

### LINKED ISSUE #1072: Make links shared in chats clickable
Simple as that, currently they are rendered as plain text.

undercase:
Fixes #1072. Uses [react-native-hyperlink](https://github.com/obipawan/react-native-hyperlink) to handle detecting links and making them clickable cross-platform.


undercase:
The module I used to detect hyperlinks and make them clickable doesn't have type declarations. It works well and considering how simple its implementation is, I think it's reasonable to just declare the module without going through and defining our own types for it for the time being.

PR #1092, opened 2019-05-25
editorconfig, prettier, eslint

balupton:
adds editorconfig, eslint, and prettier

disables `no-undefined` and `default-case` eslint rules, as they are not compatible with our typescript conventions

ignores `react/no-string-refs` as that will need to be fixed properly at a later date


package.json
```diff
@@ -6,7 +6,7 @@
     "start": "node node_modules/react-native/local-cli/cli.js start",
     "test": "jest",
     "tsc": "tsc -p . --noEmit",
-    "lint": "tslint -c tslint.json --project ."
+    "lint": "eslint --fix --ext .mjs,.js,.jsx,.ts,.tsx ."
```
asutula: Ok so part of our CI is to call `yarn lint` to make sure it passes all the lint rules. Wondering if we should remove `--fix` here since the point is to test if linting passes. If a developer wants to force fixes to run, they can locally run `yarn lint -- --fix`?

balupton: Shouldn't have any impact, as the code should already be fixed by the time it reaches CI. Without --fix, then a autofixable linting error would cause a failure, which is a trivial thing causing a build failure.

balupton:
No idea why the build is still failing for https://github.com/textileio/photos/pull/1092/commits/8bf0f072dd5bfd5d2884e27309b8180774b507c7

balupton:
argh, out of date again

balupton:
remerged, hopefully build passses

balupton:
no idea why the build keeps failing

asutula:
Looks like that call to `requireNativeComponent` that we looked at yesterday:
```
yarn run v1.13.0
$ tsc -p . --noEmit
App/Components/TextileImage.tsx:59:26 - error TS2554: Expected 1 arguments, but got 2.

 59 const TextileImageView = requireNativeComponent(
                             ~~~~~~~~~~~~~~~~~~~~~~~
 60   'TextileImageView',
    ~~~~~~~~~~~~~~~~~~~~~
... 
 62   TextileImage
    ~~~~~~~~~~~~~~
 63 )
```

asutula:
Didn't we put a @ts-ignore in there?

asutula:
@balupton I just tested that method with one argument and it works fine:
```
const TextileImageView = requireNativeComponent('TextileImageView')
```

Just update it and remove the @ts-ignore

balupton:
> Didn't we put a @ts-ignore in there?

yup

PR #1094, opened 2019-05-25
Add direct message button to contacts

undercase:
- Extends the redux threads api in the application to support types, sharing, and the whitelist.
- Adds a button to the contact modal that says "Direct Message." When this button is clicked, the application searches for a direct message thread with that contact (defined as a thread with only that contact on the whitelist, with the thread type `open` and sharing `not_shared`) and navigates to it if one is found. If one isn't found it creates one. It doesn't search for a MEDIA schema since some users may create direct message threads on other devices for just messages, or just photos, and I think we should recognize these as well.


App/Redux/PhotoViewingSelectors.ts
```diff
@@ -161,6 +161,23 @@ export function getThreadThumbs(state: RootState, byAddres: string, sortBy?: 'na
...
+           thread.type === Thread.Type.OPEN
+  }
+}
+
+export function directMessageThreadExists(state: RootState, address: string): boolean {
+  return getThreads(state).some(isDirectMessageThread(address))
+}
+
+// Returns the thread object or undefined if it isn't found
+export function getDirectMessageThread(state: RootState, address: string): ThreadData | undefined {
```
asutula: It's ok (even better) to just allow TypeScript infer the return types in all these functions.

App/Containers/Contact.tsx
```diff
@@ -119,21 +157,35 @@ const mapStateToProps = (state: RootState, ownProps: NavigationScreenProps<NavPr
...
   return {
     displayName: username ? username : address.substring(0, 12),
     threadThumbs: getThreadThumbs(state, address, 'name'),
     isContact,
     removing,
-    adding
+    adding,
+    doesDirectMessageThreadExist,
+    navigateToDirectMessageThread: () => {
+      const directMessageThread = getDirectMessageThread(state, address) as ThreadData
```
asutula: This cast using `as ThreadData` is dangerous. The return value can actually be `undefined`. I realize there is other logic in this component that would prevent this code from being called unless there is a direct message thread, but it's better to just respect the types not make any assumptions. This is like a bug waiting to happen when this assumption is no longer valid.

App/Redux/PhotoViewingRedux.ts
```diff
@@ -1,20 +1,36 @@
...
+  sharePhoto?: {
+    imageId: string
+    comment?: string
+  }
+}
 
 const actions = {
   insertThread: createAction('INSERT_THREAD', (resolve) => {
-    return (id: string, key: string, name: string) => resolve({ id, key, name })
+    return (id: string, key: string, name: string, type: Thread.Type, sharing: Thread.Sharing, whitelist: ReadonlyArray<string>) => resolve({ id, key, name, type, sharing, whitelist })
```
asutula: Just curious why you're not using the `ThreadConfig` type defined above as a parameter in these actions (you're instead listing all the same properties as individual params). Maybe it's fine, just wondering.

App/Redux/PhotoViewingSelectors.ts
```diff
@@ -161,6 +161,23 @@ export function getThreadThumbs(state: RootState, byAddres: string, sortBy?: 'na
     })
 }
 
+function isDirectMessageThread(address: string): (thread: ThreadData) => boolean {
```
asutula: coool curried function!

App/Containers/Contact.tsx
```diff
@@ -119,21 +157,35 @@ const mapStateToProps = (state: RootState, ownProps: NavigationScreenProps<NavPr
...
+  const doesDirectMessageThreadExist = directMessageThreadExists(state, address)
   return {
     displayName: username ? username : address.substring(0, 12),
     threadThumbs: getThreadThumbs(state, address, 'name'),
     isContact,
     removing,
-    adding
+    adding,
+    doesDirectMessageThreadExist,
+    navigateToDirectMessageThread: () => {
```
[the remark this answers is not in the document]
undercase: Agreed, not sure what I was thinking here.

App/Redux/PhotoViewingRedux.ts
undercase: That's probably better-since some other parts of the codebase rely on these actions, and I would need to edit them as well, I decided to come back once I got the direct message feature working and refactor these. Fixing it now.

undercase:
Thanks for the feedback @asutula! I fixed the issues you mentioned and refactored some things.

App/Redux/PhotoViewingSelectors.ts
```diff
@@ -161,6 +161,23 @@ export function getThreadThumbs(state: RootState, byAddres: string, sortBy?: 'na
...
+  return (thread) => {
+    return thread.whitelist.indexOf(address) !== -1 &&
+           thread.sharing === Thread.Sharing.NOT_SHARED &&
+           thread.type === Thread.Type.OPEN
+  }
+}
+
+export function directMessageThreadExists(state: RootState, address: string) {
```
asutula: This one seems redundant. The below `getDirectMessageThread` gives us the same information. If it returns `undefined` we know the direct message thread doesn't exist.

App/Containers/Contact.tsx
```diff
@@ -107,6 +137,22 @@ class ContactModal extends React.Component<Props> {
   onAdd = () => {
     this.props.addContact()
   }
+
+  navigateToDirectMessageThread = () => {
+    // Could be strange bug where directMessageThreadExists is true but the thread is undefined
```
asutula: My other comment about getting rid of that redundant selector... That would eliminate this potential bug because the thread as well as if it exists or not would be based on the exact same information.

undercase: Awesome, that makes sense. Thanks for catching that!

undercase:
Removed the redundant selector.

PR #1096, opened 2019-05-28
Process photos using a queue and workers

### LINKED ISSUE #1073: Need photo add queue
After some testing over the weekend traveling (and taking lots of photos), I'm seeing that the app crashes when trying to process many photo adds. 

We "prepare" the photos (`prepareFiles` call) using a go routine internally (background thread), so I think all the prepares are running simultaneously. I'm guessing this uses a lot of memory and the app is being killed because it's using too much memory. This is especially bad when it happens during a background fetch session because, on iOS, the background fetch completion handler is never called and iOS penalizes the app by reducing the number of future background fetch sessions.

asutula:
We want some control over how concurrency is used to process camera roll photos, adding them to the local Textile node. This allows us to balance speed and memory consumption. 

Closes #1073

### DIFF
--- a/App/features/photos/actions.ts
+++ b/App/features/photos/actions.ts
@@ -12,6 +12,10 @@ export const updateLastQueriedTime = createAction('photos/UPDATE_LAST_QUERIED_TI
   return (epochSeconds: number) => resolve({ epochSeconds })
 })
 
+export const photoProcessingBegan = createAction('photo/PHOTO_PROCESSING_BEGAN', (resolve) => {
+  return (id: string) => resolve({ id })
+})
+
 export const photoPrepared = createAction('photos/PHOTO_PREPARED', (resolve) => {
   return (id: string, preparedFiles: IMobilePreparedFiles) => resolve({ id, preparedFiles })
 })
--- a/App/features/photos/reducer.ts
+++ b/App/features/photos/reducer.ts
@@ -46,8 +46,14 @@ export default combineReducers<PhotosState, PhotosAction>({
     switch (action.type) {
       case getType(actions.queryCameraRoll.success): {
         return action.payload
-          .map((photo): ProcessingPhoto => ({ photo, state: 'preparing' }))
+          .map((photo): ProcessingPhoto => ({ photo, state: 'pending' }))
           .reduce((accum, processingPhoto): ProcessingPhotos => ({ ...accum, [processingPhoto.photo.assetId]: processingPhoto }), state)
+      }
+      case getType(actions.photoProcessingBegan): {
+        const { id } = action.payload
+        const processingPhoto = state[id]
+        const updated: ProcessingPhoto = { ...processingPhoto, state: 'preparing' }
+        return { ...state, [id]: updated }
       }
       case getType(actions.photoPrepared): {
         const { id, preparedFiles } = action.payload
--- a/App/features/photos/sagas.ts
+++ b/App/features/photos/sagas.ts
@@ -1,4 +1,5 @@
-import { all, take, select, call, put, fork, takeEvery } from 'redux-saga/effects'
+import { buffers, channel, Channel } from 'redux-saga'
+import { all, take, select, call, put, fork, takeEvery, cancelled } from 'redux-saga/effects'
 import { getType, ActionType } from 'typesafe-actions'
 import { requestLocalPhotos, LocalPhotoResult } from '@textile/react-native-camera-roll'
 import Textile, { IThread, IMobilePreparedFiles, IFilesList } from '@textile/react-native-sdk'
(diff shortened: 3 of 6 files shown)


App/features/photos/actions.ts
```diff
@@ -12,6 +12,10 @@ export const updateLastQueriedTime = createAction('photos/UPDATE_LAST_QUERIED_TI
   return (epochSeconds: number) => resolve({ epochSeconds })
 })
 
+export const photoProcessingBegan = createAction('photo/PHOTO_PROCESSING_BEGAN', (resolve) => {
+  return (id: string) => resolve({ id })
+})
```
asutula: This action is used to move the photo from `pending` to `preparing` state

App/features/photos/reducer.ts
```diff
@@ -46,9 +46,15 @@ export default combineReducers<PhotosState, PhotosAction>({
     switch (action.type) {
       case getType(actions.queryCameraRoll.success): {
         return action.payload
-          .map((photo): ProcessingPhoto => ({ photo, state: 'preparing' }))
+          .map((photo): ProcessingPhoto => ({ photo, state: 'pending' }))
```
asutula: Photos are now added to the persistent redux state in a 'pending' state.

App/features/photos/sagas.ts
```diff
@@ -22,16 +23,49 @@ function * toggleStorage(action: ActionType<typeof PreferencesActions.toggleStor
   }
 }
 
-function * nodeOnline() {
-  while (yield take([getType(TextileEventsActions.nodeOnline), getType(TextileEventsActions.stopNodeAfterDelayCancelled)])) {
+function * bootstrapPhotoProcessing() {
+  const QUEUE_CONCURRENT = 1
```
asutula: For now, processing one photo at a time.

App/features/photos/sagas.ts
```diff
@@ -22,16 +23,49 @@ function * toggleStorage(action: ActionType<typeof PreferencesActions.toggleStor
...
+  const { watcher, addTaskChannel } = yield createQueue(photoHandler, QUEUE_CONCURRENT)
```
asutula: This creates our processing queue and workers, provides a channel to add new photo processing tasks to

App/features/photos/sagas.ts
```diff
@@ -22,16 +23,49 @@ function * toggleStorage(action: ActionType<typeof PreferencesActions.toggleStor
...
+  yield fork(watcher)
+  yield call(processPending, addTaskChannel)
```
asutula: Before we allow any camera roll querying to happen, we want to add any still-pending photos to the queue first.

App/features/photos/sagas.ts
```diff
@@ -22,16 +23,49 @@ function * toggleStorage(action: ActionType<typeof PreferencesActions.toggleStor
...
+  yield fork(queryForNewPhotos, addTaskChannel)
```
asutula: Then we fork off our saga that queries for new camera roll photos

App/features/photos/sagas.ts
```diff
@@ -22,16 +23,49 @@ function * toggleStorage(action: ActionType<typeof PreferencesActions.toggleStor
...
+}
+
+function * photoHandler(payload: { id: string }) {
```
asutula: This is our pipeline for processing a single photo. It pauses until the node is online and then prepares, adds, and cleans up the photo.

App/features/photos/sagas.ts
```diff
@@ -41,48 +75,40 @@ function * queryForNewPhotos() {
...
       yield put(actions.queryCameraRoll.success(results))
       yield put(actions.updateLastQueriedTime(currentRefresh))
-      for (const localPhotoResult of results) {
-        yield fork(preparePhoto, localPhotoResult.assetId)
-      }
+      yield all(
+        results.map((result) =>
+          put(addTaskChannel, { payload: { id: result.assetId } })
+        )
+      )
```
asutula: Here we now just add any new photo tasks to the queue instead of processing them directly

App/features/photos/sagas.ts
```diff
@@ -136,11 +162,45 @@ function * watchForLoadPhotosRequests() {
   }
 }
 
+function* createQueue(handle: (...args: any[]) => IterableIterator<any>, concurrent = 1) {
```
asutula: Here's the fancy redux-saga code that created a concurrent queue/worker setup

App/screens/photos.tsx
```diff
@@ -113,7 +113,8 @@ class Photos extends Component<Props> {
         />
       )
     } else {
-      const { photo, error } = row.item.processingPhoto
+      const { photo, state, error } = row.item.processingPhoto
+      const color = state === 'pending' ? 'white' : 'blue'
```
asutula: Display a white circle on `pending` photos, blue circle on photos that are currently being processed.

undercase: Considering that `color` is another style attribute used to change the font color, why not just directly call this variable `backgroundColor`.

asutula: Yea totally agree. This is temporary code that will be thrown out as soon as we make the real UI here, so wasn't worried too much about the nice details like that ;)

PR #1098, opened 2019-05-29
Display loading component while loading persisted data

asutula:
We were previously not instructing redux-persist to display any view while loading the persisted data. This PR tells redux-persist to display our `Loading` component with a red spinner while it's loading the persisted data.

I also updates the `StatusCheck` screen (which blocks any UI from rendering until the node is started) to display a blue spinner.

The idea is we can know, while the spinners are displaying during app startup, which is taking a long time.

### DIFF
--- a/App/Components/Loading.tsx
+++ b/App/Components/Loading.tsx
@@ -1,0 +1,20 @@
+import React from 'react'
+import { View, ActivityIndicator, ActivityIndicatorProps } from 'react-native'
+
+import { color } from '../styles'
+
+interface Props extends ActivityIndicatorProps {
+  backgroundColor?: string
+}
+
+const loading = (props: Props) => {
+  const backgroundColor = props.backgroundColor ? props.backgroundColor : color.screen_primary
+  const indicatorProps: ActivityIndicatorProps = { size: 'large', ...props }
+  return (
+    <View style={{ flex: 1, alignContent: 'center', justifyContent: 'center', backgroundColor }}>
+      <ActivityIndicator {...indicatorProps} />
+    </View>
+  )
+}
+
+export default loading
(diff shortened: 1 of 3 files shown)


App/Components/Loading.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import { View, ActivityIndicator, ActivityIndicatorProps } from 'react-native'
+
+import { color } from '../styles'
+
+interface Props extends ActivityIndicatorProps {
+  backgroundColor?: string
+}
+
+const loading = (props: Props) => {
```
asutula: Just pulled out the loading screen into a reusable component.

App/Containers/App.tsx
```diff
@@ -27,7 +29,7 @@ class App extends Component {
   render() {
     return (
       <Provider store={store}>
-        <PersistGate loading={undefined} persistor={persistor}>
+        <PersistGate loading={<Loading color={color.brandRed} />} persistor={persistor}>
```
asutula: Tell `redux-persist` to display the `Loading` component

App/Containers/StatusCheck.tsx
```diff
@@ -40,9 +42,7 @@ class StatusCheck extends React.Component<Props, {}> {
       )
     } else {
       return (
-        <View style={{ flex: 1, alignContent: 'center', justifyContent: 'center' }}>
-          <ActivityIndicator size='large' />
-        </View>
+        <Loading color={color.brandBlue} />
```
asutula: Display a blue spinner while we wait for the node to start. @andrewxhill see above on line 24 and 26 how the node has to be started before it will navigate to either the onboarding or main app navigation (from a cold start, that is)

asutula: If the node isn't started, it either displays the loading spinner or an error message

asutula:
@andrewxhill really curious to see how this behaves on your phone. I'm hoping the black screen will no longer be displayed at all, or if at all, very quickly before the persistence loading component is displayed.

PR #1099, opened 2019-05-29
Prevent users from liking a photo more than once.

### LINKED ISSUE #958: Incorrect number of 'likes' on a single Photo
**Describe the bug**
A user noticed that some of the pictures in a group have more likes on them than there are people in the group.

**To Reproduce**
1. Create a group
2. Share invite link and have peers join
3. Create post, have member 'like' the post
4. Check post 'likes', count is greater than number of participants

**Expected behavior**
This should only ever be, at most, equal to the number of participants in the group

undercase:
Fixes #958. The framework is there to log errors to the state, but the reducer doesn't because I can think of no good place to display them/clear them.


App/Sagas/TextileSagas.ts
```diff
@@ -150,11 +150,17 @@ export function * backgroundLocationPermissionsTrigger() {
...
 }
 
-export function * addPhotoLike(action: ActionType<typeof UIActions.addLikeRequest>) {
+export function * addPhotoLike(action: ActionType<typeof UIActions.addLike.request>) {
   const { blockId } = action.payload
   try {
-    yield call(Textile.likes.add, blockId)
+    const likingPhotos = yield select(UISelectors.likingPhotos)
+    // If photos is already being liked, don't let the user like it again
+    if (Object.keys(likingPhotos).indexOf(blockId) !== -1) {
```
asutula: Maybe I'm reading this wrong, but....
1. You get the list of block ids in the process of being liked
2. You check if that list already contains the block id trying to be liked now
3. If the new block id isn't in the list, `indexOf` should return -1
4. But you're checking that it returns something other than -1

undercase: Okay, this is totally wrong, good catch but it begs the question-How was it working for me locally?? I must have already populated the state before setting this or something.

undercase:
Totally, I figured that there could be multiple places in the app to like a photo, so I centralized the check in the saga. My rational was also that the “like” button itself is a heart icon, and in other apps with that UI there’s no “disabled” state (like in instagram, the heart isn’t “disabled” at all as far as I can tell), so I didn’t even try to implement it. I’ll look into it more.

App/Sagas/TextileSagas.ts
asutula: 🤷‍♂

asutula:
Yea, in Instagram, you can unlike a photo by tapping again. We don't have that. What we do have is:

1. You can like a photo
2. We don't currently try to display an error if one happens during the like adding (unlikely)

So, I guess the best UX for what we do have is to just make sure the user doesn't feel like they need to tap the like button again while the processing is happening. The easiest way to do that would be to just render the "liked" state of the heart if the like is processing OR complete. That way it just changes to look liked immediately, and you can't tap it again.

undercase:
Totally, that makes sense. Fixed the issues you mentioned and configured it to show photos as liked during the pending request as well.

PR #1105, opened 2019-05-30
Don't display duplicate join items

### LINKED ISSUE #1044: Hide Join notifications with it's just another device from existing user
Joins are showing up in the Group feeds

undercase:
Fixes #1044. Adds a filter to the `feedItems` selector that removes join items from the same user after the first one. The only issue I foresee is that this will suppress items indicating a user actually joining a group more than once (i.e. leaving and then rejoining). Messenger just shows every time a user leaves/joins. If we want to go that route, then I need to know how to tell an actual unique join item from one that just represents a new device joining for the same user.


asutula:
This looks ok to merge if we want to, if we're fine with the issue that @undercase describes. It's also an opportunity to re-think how we are doing join/leave blocks. Do we want to look at doing something like having two types of join/leave blocks: account and peer? Is that even possible? If so, an app like Photos would probably just ignore peer joins/leaves and only display them in terms of account. Thoughts @sanderpick ?

undercase:
So do we want to go ahead and merge this as a temporary fix @sanderpick @asutula?

sanderpick:
fine by me!

sanderpick:
How about this one?

asutula:
Yea, I guess i'd be fine with this for now. It's a very temporary solution that will break as soon as we move to paged data. We should make some plans for the long term solution i asked you about above. @undercase this one has some conflicts too, let's get them resolved and look at merging this.

asutula:
@undercase should we get this guy up to date and merged?

undercase:
Definitely, will do this today.

undercase:
This isn't good long-term, so I'm closing it. Let's solve this on the textile side instead of the client side.

PR #1116, opened 2019-06-03
Update yarn.lock for current package.json

asutula:
Pulled the latest master and ran a `yarn install` which resulted in these updates to `yarn.lock`. I'm assuming that the recent PR from @balupton didn't include the update because he probably used `npm` locally.


balupton:
guilty as charged

PR #1120, opened 2019-06-04
Consolidate contacts screen and add contacts screen

### LINKED ISSUE #1108: Consolidate contact list and search
The two screens are confusing and could be combined into one. I suppose that in it's normal state, it should list all your account contacts, and in search mode, it should list search results grouped by type (account contacts, network search results, address book)

undercase:
Fixes #1108. Uses the same rendering logic as in `AddContact.tsx`, but in `Contacts.tsx`.


undercase:
@asutula Fixed the placeholder text!

PR #1124, opened 2019-06-05
List own cafes

### LINKED ISSUE #1111: View list of own cafes
This should exist in the settings or wallet/account screens somewhere.

undercase:
Fixes #1111. Adds a button to the settings drawer that displays a simple list of the addresses of the cafes one is connected to.

### DIFF
--- a/App/Containers/Cafes.tsx
+++ b/App/Containers/Cafes.tsx
@@ -3,16 +3,8 @@ import { connect } from 'react-redux'
 import { NavigationScreenProps } from 'react-navigation'
 import { RootState } from '../Redux/Types'
 
-import {
-  Text,
-  ScrollView,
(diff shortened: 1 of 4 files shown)


asutula:
Think I wanna hit the pause button on this one for now, work with you on some UI ideas in mock ups, then re-implement this UI with whatever we decide on. 

This technically checks the box for displaying the list of cafes, but I think we can do it in a beautiful way that adds to the UI/UX we want to create for Textile. These sorts of features can go a long way in providing the feel and context of the system we're creating, so let's take it there.

undercase:
For sure, this was quick & dirty. Would love to make something more elegant and insightful.

asutula:
Closing this because of all the new work on cafes

PR #1126, opened 2019-06-05
Move node event handling to a saga

asutula:
Ok @andrewxhill, I think this will help and hopefully solve the empty UI problem you've been seeing. I verified that it does fix that problem in the case of running the dev build on a device from Xcode and reloading the JS.

Main thing here is a change in the timing of things.

Previously, we were initializing the Textile node in a saga, but subscribing to it's events in a class hung off the React Native UI component in `App.tsx`. If the UI took a little longer to load, the Textile node start event could fire before the event subscription was set up. The start event is what triggers the loading of data for the UI, hence the empty UI.

This new setup subscribes to node events using a redux-saga `eventChannel`, and that channel is consumed in a saga that is initialized before the saga that initializes the Textile node. Therefore, the subscription is always set up before the node is even initialized and things happen in the correct sequence.


andrewxhill:
ah yeah, nice. looks good

App/Sagas/TextileEventsSagas.ts
```diff
@@ -19,11 +34,140 @@ function displayNotification(message: string, title?: string) {
...
+    | DeviceLogsAction
+    | NotificationsAction
+    | TextileEventsAction
+  >(emitter => {
+    const subscriptions: EventSubscription[] = []
+    subscriptions.push(
+      Textile.events.addThreadUpdateReceivedListener(update => {
+        const { type_url } = update.payload
+        if (
+          type_url === '/Text' ||
```
undercase: Might it be better to put these type strings into a list, like `refreshTypes`, and check if `type_url` is in that list?

App/Sagas/TextileEventsSagas.ts
```diff
@@ -19,11 +34,140 @@ function displayNotification(message: string, title?: string) {
...
+        ) {
+          emitter(groupActions.feed.refreshFeed.request({ id: update.thread }))
+          // FIXME: This is a hack. We need to examine the thread id and dispatch one or the other.
+          // Or this needs to send the whole Thread or at least the addition of key
+          emitter(photosActions.refreshPhotos.request(undefined))
+        }
+
+        // TODO: remove this if needed
+        if (
+          type_url === '/Comment' ||
```
undercase: Same as above.

App/Sagas/TextileEventsSagas.ts
asutula: Possibly something like that. It would take multiple lists because we're matching against a unique set each time. But anyway, this code was copy/pasted from another file, and I want to keep this PR focused on the task at hand. We can revisit how to deal with these protobuf `Any` types as another task (the answer is probably expanding the callbacks in the SDK layer so we don't pass `Any` objects in here).

PR #1135, opened 2019-06-07
SDK 1.2.11 (go-textile 0.2.6)

asutula:


asutula:
This only updates the ios side of the sdks since bintray is down. Will try android again in the morning.

PR #1138, opened 2019-06-07
uses a flatlist instead of scrollview-textarea for android log

andrewxhill:
scrollview approach works well on iOS so i didn't want to remove it. at the same time, it doesn't work on android because of reason linked here, https://github.com/textileio/photos/issues/1136. so for android it uses a flatlist and allows row-based copy by touch.


andrewxhill:
oop, didn't realize i never clicked merge. going to do so now

PR #1151, opened 2019-06-08
detects the error i see "unauthorized" and triggers a cafe session refresh

andrewxhill:
Signed-off-by: andrew <andrew@textile.io>

### DIFF
--- a/App/Sagas/TextileEventsSagas.ts
+++ b/App/Sagas/TextileEventsSagas.ts
@@ -199,6 +199,9 @@ export function* refreshMessages() {
       yield call(logNewEvent, 'refreshMessages', action.type)
     } catch (error) {
       yield call(logNewEvent, 'refreshMessages', error.message, true)
+      if (error.message === 'unauthorized') {
+        yield put(accountActions.refreshCafeSessionsRequest())
+      }
     }
   }
 }


undercase:
I also need this-started getting the unauthorized error on my iPhone.

asutula:
Andrews phone was in a state where his refresh token was expired, so he was basically stuck because the app doesn’t currently deal with re registering the cafe in that case. 

Maybe you’re in the same situation, but I doubt it. You can check it the status of your sessions by going to the settings menu in verbose mode and tapping on Cafes. You can refresh your session there if necessary. 

I’m considering making an issue to have the app deal with cafe re registration in the case of expired refresh tokens, but all of this stuff is being moved to the native sdks in the next update.

undercase:
Just did this, and that was the issue. They were both expired.

sanderpick:
@asutula sounds good... we can tag on a `ReRegisterCafe` method that should help here too. No need to re-announce to contacts.

asutula:
Were you able to refresh the token @thomas or are you stuck as well?

undercase:
I’m stuck.

andrewxhill:
what's going on here?

asutula:
There’s a more general fix for refreshing expired session in the photos app and in the sdk layer. The photos app fix is the quicker short term fix and I’ll have a pr out today. Longer term sdk fix is wip and not top priority

asutula:
Fixed in https://github.com/textileio/photos/pull/1214

PR #1152, opened 2019-06-08
Clean up redux

asutula:
Removes some data that is no longer used and some other large data that should have never been in redux. Hoping this will improve load times a little.


App/Redux/RootReducer.ts
```diff
@@ -16,9 +17,356 @@ import { reducer as textileEventsReducer } from './TextileEventsRedux'
 import { groupReducer } from '../features/group'
 import { photosReducer } from '../features/photos'
 
-export default combineReducers({
+const migrations: MigrationManifest = {
```
asutula: Easier to put the persist config stuff with the reducer it belongs to

App/features/group/reducer.ts
```diff
@@ -17,9 +19,19 @@ export interface GroupState {
 
 export type GroupAction = ActionType<typeof actions>
 
-export default combineReducers<GroupState>({
+const persistConfig: PersistConfig = {
```
asutula: Using a nested persist config to control what is being persisted in a nested reducer where some things need to be persisted but not others.

asutula:
Added an option to the camera roll photos screen to clear out all processing photo items. Errored out ones build up over time, and i'm now very aware and cautious of persistent data that can grow boundlessly.

asutula:
Ok, this cleaning up of redux persist took @andrewxhill's sample data he gave me from 972KB of persisted data to 80KB. It didn't have as much an impact on startup time as I had hoped, but it's good to have cleaned up.

General new rule: Watch out for persisted redux data that can grow without bound. It should be avoided.

App/features/photos/reducer.ts
```diff
@@ -21,7 +23,14 @@ export interface PhotosState {
 
 export type PhotosAction = ActionType<typeof actions>
 
-export default combineReducers<PhotosState, PhotosAction>({
+const persistConfig: PersistConfig = {
+  key: 'photos',
+  storage: AsyncStorage,
+  whitelist: ['queryData', 'processingPhotos'],
```
asutula: Another nested persist config that persists the camera roll last queried time and processing photos, but not the camera roll photos queried from the Textile node.

App/screens/photos.tsx
```diff
@@ -197,7 +227,8 @@ const mapStateToProps = (state: RootState): StateProps => {
 
 const mapDispatchToProps = (dispatch: Dispatch<RootAction>): DispatchProps => ({
   queryPhotos: () => dispatch(photosActions.queryCameraRoll.request()),
-  refreshPhotos: () => dispatch(photosActions.refreshPhotos.request(undefined))
+  refreshPhotos: () => dispatch(photosActions.refreshPhotos.request(undefined)),
+  clearProcessingPhotos: () => dispatch(photosActions.clearProcessingPhotos())
```
asutula: Option from the camera roll photos screen to clear out processing items since any failed ones (which there are plenty of at this point) pile up over time.

asutula:
FYI... The migration that runs in here will result in any processing camera roll photos to be cleared out from the photos screen/redux.

PR #1157, opened 2019-06-09
removes force cache from avatar requests

andrewxhill:
Signed-off-by: andrew <andrew@textile.io>


asutula:
`The data for the URL will be loaded from the originating source. No existing cache data should be used to satisfy a URL load request.`

PR #1166, opened 2019-06-13
Send invite on direct message

### LINKED ISSUE #1129: Need to sort out "send message" UX on the receiving end
Currently, there is no UX at all. If you send a message to a contact, they will never know it happened, they will never see the message. 

The UX we'd want could be what you'd expect... The user is notified of the message and can chose to start participating or decline. What would that take using the tools in our Textile toolbox? Is this just as simple as: In addition to creating the thread, we need to send an invite as well?

undercase:
Fixes #1129. Features:

- Extends `CreateThreadModal` to support configuration parameters (`ThreadConfig`).
- Opens up a `CreateThreadModal` when creating a direct message thread so that the user can choose its initial name.
- Sends an invite to the contact when a direct message thread is created.


App/Components/CreateThreadModal.tsx
```diff
@@ -53,8 +69,19 @@ class CreateThreadComponent extends React.Component<
       }
       this.setState({ submitted: true })
       this.props.completeScreen(this.state.value)
+      const threadConfig = {
+        type:
+          this.props.type !== undefined ? this.props.type : Thread.Type.OPEN,
+        sharing:
+          this.props.sharing !== undefined
+            ? this.props.sharing
+            : Thread.Sharing.SHARED,
```
andrewxhill: couldn't this just be `this.props.sharing || Thread.Sharing.SHARED`

App/Components/CreateThreadModal.tsx
```diff
@@ -53,8 +69,19 @@ class CreateThreadComponent extends React.Component<
...
+        whitelist:
+          this.props.whitelist !== undefined ? this.props.whitelist : [],
+        name: this.state.value
+      }
```
andrewxhill: this constructing of the ThreadConfig in the view Component might set of Aaron alarms, where he might suggest that it get it's own Saga to handle the transformation before attempting to create the thread. I'm okay with it for now though! :)

App/Containers/Contact.tsx
```diff
@@ -101,6 +104,13 @@ class ContactModal extends React.Component<Props> {
     }
   }
 
+  constructor(props: Props) {
```
andrewxhill: i've never been sure the difference between setting the state in a constructor versus just setting it in the root tree... neither our lint nor compiler have a preference, so neither do I, yet.

App/Redux/PhotoViewingRedux.ts
```diff
@@ -3,14 +3,14 @@ import Config from 'react-native-config'
 
 import { IFiles, Thread } from '@textile/react-native-sdk'
 
-interface ThreadConfig {
+export interface ThreadConfig {
```
andrewxhill: yep, I can hear aaron's alarms for sure, given that this now needs to be exported instead of keeping the logic tight in this file.

App/Components/CreateThreadModal.tsx
undercase: It used to be this, actually, but when `this.props.sharing` is equal to `Thread.Sharing.NOT_SHARED`, which is the enum value 0, it evaluates to false and the logic breaks down, so I had to explicitly check it against undefined.

App/Containers/Contact.tsx
undercase: Hmm I'm not sure, I've always set it in the constructor just because that's what the React docs do.

App/Components/CreateThreadModal.tsx
andrewxhill: ah hahah! aaron was just talking about this issue this morning. good call.

App/Containers/Contact.tsx
andrewxhill: 👍

App/Components/CreateThreadModal.tsx
andrewxhill: i meant to say, 'Good, you passed my test'.

undercase: haha yeah it's a really pernicious bug! took me forever to figure out why it was selectively failing haha.

asutula:
Sorry this one got a little stale. Let's get it updated, tests passing, and merged!

undercase:
@andrewxhill I removed the dependency on `ThreadConfig` in `CreateThreadModal` and it's no longer exported. This should be ready for further review & merge.

undercase:
@asutula  `CreateThreadModal` will send invites once the user submits from that screen. I didn't manually add in the code to send an invite, I connected the button to `CreateThreadModal`, which handles it for me.

undercase:
Anything I can do to get this ready for merge?

sanderpick:
@asutula can we get this in? I think it's currently broken in the app when you "send a message", since no invite is sent and there is no way to add after the fact. My guess is this is a pretty common entry point to sharing since it's so intuitive.

asutula:
Yea for sure. @undercase you want to get this thing up to date with master?

PR #1174, opened 2019-06-20
Use latest SDK

asutula:
Cleans up feed related code now that we don't have to unmarshall protbuf `Any` from `FeedItem`.


App/Sagas/TextileEventsSagas.ts
```diff
@@ -46,53 +49,57 @@ function nodeEvents() {
...
+      Textile.events.addThreadUpdateReceivedListener(
+        (threadId, feedItemData) => {
+          if (
+            feedItemData.type === FeedItemType.Text ||
+            feedItemData.type === FeedItemType.Comment ||
+            feedItemData.type === FeedItemType.Like ||
+            feedItemData.type === FeedItemType.Files ||
+            feedItemData.type === FeedItemType.Ignore ||
+            feedItemData.type === FeedItemType.Join ||
+            feedItemData.type === FeedItemType.Leave
```
andrewxhill: super

asutula: Yea, we're not using it in Photos, but theis `FeedItemData` type is especially powerful when you switch on it's `type` property... the compiler resolves the type of the associated `value` correctly.

PR #1175, opened 2019-06-24
Add Cafes UI

undercase:
This is a draft pull request. The user interface code is finished, and it needs to be connected to the backend before this will be ready to merge.


undercase:
These are completely done (as in the UI is there, it is connected to the backend, and everything works):

- Register with a cafe (or choose not to) on onboarding.
-- Show list of Textile Cafes, and mark the two nearest cafes as *Recommended* in the UI.
-- Give user the option to enter their own cafe URL and token.
- View list of cafes from settings.
- View individual cafe, showing its peer id, list of services (hardcoded for now), and a deregister button.
-- Deregister from cafe and go back to list of cafes screen when the user clicks the deregister button.
- Register with a new cafe from settings.

Tomorrow, I am going to polish this code (pull out some reused code into models and make sure the error messages render correctly) and write some tests for the new reducer, then open this PR. After this, I'm going to add some code to the contact screen to highlight shared cafes (cafes both the user and the contact are registered with).

sanderpick:
Nice! Couple things:
- You can do tests in a separate PR as it would be good to have this to play with in beta ASAP
- We should add the new cafe to this list, but it's not part of the location based recommendation... so maybe we just skip that and make them choose. The only option for Europe should be the Azure one.

I've slowly come to the realization that we really ought to have just three larger prod cafes that a user can manually choose from:
  - EU West
  - US East
  - AS South (Singapore)
This can come after camp. We can move the location-based recommendation test to Cloudflare as well.

App/Containers/ChooseCafe.tsx
```diff
@@ -0,0 +1,203 @@
...
+        ? this.props.registeringCafes[url].error
+        : undefined
+    const registering = registrationStarted && !error
+    const buttonDisabled = !this.state.selected || registering
+    return (
+      <SafeAreaView style={Container}>
+        <Text style={Header}>Choose a Cafe</Text>
+        <Text style={Subheader}>
+          Cafes are trustless, always-on nodes that assist the peer network.
+          Don&apos;t select a cafe to proceed without registering with one.
```
asutula: This sentence doesn't make sense to me

App/Containers/ChooseCafe.tsx
```diff
@@ -0,0 +1,203 @@
...
+        </Text>
+        <CafesList
```
asutula: RN is displaying a warning about non-unique `key`s in the list

App/Components/CafeListHeader.tsx
```diff
@@ -0,0 +1,72 @@
...
+  render() {
+    return (
+      <View style={Wrapper}>
+        <TouchableOpacity onPress={this.props.onPress}>
+          <View style={Container}>
+            <View style={IconContainer}>
+              <Icon name="search" size={size._032} color={color.grey_3} />
+            </View>
+            <View style={InfoStyle}>
+              <Text style={LabelStyle}>Search for a Cafe by Peer ID</Text>
```
asutula: This says "search", but it's actually how you enter a custom cafe. Cafe search doesn't seem like an important use case to me, so i'd just update the working to be "enter your own cafe information"

App/Components/CafePeerIdModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+}
+
+type Props = OwnProps
+
+interface State {
+  peerId: string
+  token: string
+}
+
+export default class CafePeerIdModal extends Component<Props, State> {
```
asutula: This screen needs some `SafeAreaView` love. Looks real bad on iPhone X

App/Containers/Cafes.tsx
```diff
@@ -0,0 +1,75 @@
...
+      headerLeft,
+      headerTitle: 'Registered Cafes',
+      headerRight
+    }
+  }
+
+  render() {
+    return (
+      <SafeAreaView>
+        <FlatList
```
asutula: This flatlist isn't flexing to fill the whole screen. If the number of items is less than the height of the screen, the flatlist is only as tall as the items.

App/Containers/RegisterCafe.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+interface State {
+  selected?: {
+    url: string
+    token: string
+  }
+  peerIdModalIsVisible: boolean
+}
+
+class RegisterCafe extends Component<Props, State> {
```
asutula: When I navigate to settings > cafes and tap the button to add a new cafe, I select a cafe to add and tap the arrow button at the bottom of the screen. The cafe is registered, but nothing happens, i just stay on that screen. It's confusing.

sanderpick:
Re: multiple cafes, agreed @asutula, no need for now... but we should retain the ability to do so when we have, for example, a push notification cafe. You'd have one for storage, one for notifications. The flow may be different though... you'd choose a cafe to "fulfill" that service. So, I don't think you'd ever be selecting multiple from a list at the same time.

undercase:
A couple of questions:

- Which EU cafe is the "azure" one?

- Right now, you can register with one cafe at a time (the list doesn't support multi select). I left in the ability to add multiple cafes because @sanderpick mentioned we'll want that down the road. For now, would hiding the + button to add a new cafe when the user is already registered with one solve that issue?

App/Containers/RegisterCafe.tsx
undercase: It fades the button when the registration is still processing, and I just fixed it so it will go back to the cafes screen once registration is finished.

App/Containers/Cafes.tsx
undercase: Fixed.

App/Containers/ChooseCafe.tsx
undercase: Oops, my bad, fixed the list.

App/Containers/ChooseCafe.tsx
undercase: I wasn't quite sure what copy to put here, I tried to throw something together. Do you have any pointers for me about how to describe cafes?

PR #1177, opened 2019-06-27
SDK 2.1.3

asutula:
- Pull in latest rn sdk
- Fix bug where cafe urls were being used to register instead of peer id
- Fix bug where list of registered cafes wasn't visible


asutula:
Going in

PR #1178, opened 2019-06-27
BIG cafes

asutula:
- Displays the three big cafes as only options
- Removes recommended cafes code
- Gets add custom cafe working


App/Components/CafesList.tsx
```diff
@@ -1,97 +1,34 @@
...
 }
 
-interface StateProps {
-  recommended?: DiscoveredCafes
-}
-
-interface DispatchProps {
-  requestRecommendedCafes: () => void
-}
+const cafesBase64 = Config.RN_TEXTILE_CAFES_JSON
```
asutula: This value is read from the CircleCI "context" (both beta and production). It is a base64 encoded version of a json data structure that is represented by `Cafe[]`. I've stored the current valid json in 1Password as `production-cafes.json`

sanderpick: Sweet. Does the beta app just have one value then?

asutula: Currently, it's the same for prod and beta and you can manually enter the info for beta. But, I should make them different yea?

sanderpick: I think so yeah

asutula: Ok doing that

asutula: Ok the beta release will list dev and beta cafes

asutula:
Ok g2g here i think

PR #1190, opened 2019-07-01
Ran pod install

asutula:
Didn't catch this in your previous PR @undercase, but FYI, if we add any ios dependency via Podfile, you have to run `pod install` and commit the results.


undercase:
Noted! Thanks for clearing that up.

PR #1192, opened 2019-07-02
General use ListItem component

asutula:
The number of one-off list item components was getting a little out of hand, so wanted to make it easy to just reuse `ListItem` whenever you need things in a list. This adds a optional `showDisclosure`, `selecting`, `selected`, `onSelect`, and `disabled` props, and cleans up the code. I updated all the cafe-related lists to use it and deleted those one-off implementations.


asutula:
Cleaned up some wording and the UI for triggering adding of a custom cafe as well

asutula:
This is going in to cut down on future conflicts @undercase

undercase:
This is awesome, thank you.

PR #1199, opened 2019-07-04
RN SDK 3.0

asutula:
Uses new initialize API, sets us up for new UI and flow to allow user to initialize with existing account.


App/Config/app-config.ts
```diff
@@ -0,0 +1,7 @@
+import FS from 'react-native-fs'
+
+const config = {
+  textileRepoPath: `${FS.DocumentDirectoryPath}/textile-go`
```
asutula: The sdk now requires that the host app pass in the repo path. More flexible that way, makes sense, etc etc.

asutula: Just storing our value for repo path in a centralized location where any code can access it

App/Sagas/TextileEventsSagas.ts
```diff
@@ -196,15 +197,21 @@ function* handleNodeEvents() {
...
+    if (!initialized) {
+      const phrase = yield call(
+        Textile.initializeCreatingNewWalletAndAccount,
+        AppConfig.textileRepoPath,
+        verbose,
+        true
+      )
       yield put(accountActions.setRecoveryPhrase(phrase))
     }
+    yield call(Textile.launch, AppConfig.textileRepoPath, verbose)
```
asutula: Above is an example of using the new init api to replicate the functionality that existed before (just always create a new account). This will be changed with a fork in the logic depending on if the user chooses to create a new account or sync an existing account.

asutula: 1. Always first check if Textile is already initialized
2. If not, initialize Textile, either creating a new wallet and account or by specifying an existing account seed.
3. Always launch Textile

asutula:
Post-merge feedback welcome.

undercase:
So far so good, this makes sense to me. I'll comment if anything comes up.

PR #1206, opened 2019-07-12
RN SDK 3.0.1, go-textile 0.6.2

asutula:
This also adds a check that each thread contains at least one item in `headerBlocks[]` and if not, marks the thread as invalid. The UI displays an alert icon, that when tapped, displays a message that the thread may not behave correctly and should be deleted.


asutula:
Screen Shot 2019-07-12 at 12 20 55

App/Sagas/PhotoViewingSagas.ts
```diff
@@ -138,7 +139,9 @@ export function* refreshThreads(
        */
       const isSharedThread = thread.key.indexOf('textile_photos-shared') === 0
       if (isSharedThread) {
-        yield put(PhotoViewingActions.insertThread(thread))
+        const valid = thread.headBlocks.length > 0
```
asutula: Here is the check that determines if the alert is shown to the user. Sufficient @sanderpick ?

asutula:
Want to get this out, but feel free to add comments we can address later.

PR #1211, opened 2019-07-18
Add account seed and QR code screens

### LINKED ISSUE #1209: Ability to view/copy account seed

### LINKED ISSUE #1210: Ability to view QR code

undercase:
# The Screens

## Account Seed
image

## QR Code
image

# Updates
- Pull request to address #1208.
- Add account seed to `account` redux store.
- Retrieve account seed from the SDK on startup.
- Add screens to view account seed and view account address QR code, accessible from settings. Fixes #1209. Fixes #1210.
- Hold to copy account seed on account seed screen, with an animated `COPIED` flash popup.
- Display error messages on both screens if needed.

# Questions
- Right now, the QR code screen just encodes the account address directly into the QR code. Over the past few days, I have researched deep linking and dissected the existing deep link code (for invites). I've already experimented with it locally, so I'm ready to modify the QR code screen so that it displays a link that other photos users can scan to open up that contact in their application. This would be easy to do with the existing contact screen. I would just like confirmation that is the route we want to go, and then I will implement it. I'm assuming in this that the goal of the account QR code is to make it easy to view/add contacts. If that is wrong, perhaps we should discuss this further so that I understand its intentions.


asutula:
Just to clarify, the qr code is just a shortcut to avoid having to copy your account seed on one device and somehow send it to another device to copy and paste into a text input to pair accounts during onboarding

App/Containers/AccountSeed.tsx
```diff
@@ -0,0 +1,183 @@
...
+          <Text style={ERROR}>
+            Error retrieving your account seed: {this.props.error}
+          </Text>
+        )}
+        {!this.props.seed && !this.props.error && (
+          <Text style={ERROR}>
+            Error: Your account seed has not been loaded into the redux store.
+          </Text>
+        )}
+        <Animated.View
```
asutula: There's a `react-native-easy-toast` module we use elsewhere, probably better to just stay consistent and use it.

undercase: Slipped my mind. Will switch to this.

asutula:
Hey @undercase for some reason I thought you said you already made updates based on my feedback, but I don't see any new commits... Maybe I just dreamt that up? Anyway, let me know if you need any help or input on this.

undercase:
This should be good to go.
- Consolidated screens.
- Used `toast` module.

PR #1214, opened 2019-07-19
Cafe Session Refresh/Re-register

asutula:
Ok here is a short term fix to make sure we have valid cafe sessions in the Photos app. Main changes:

- Consolidated all cafe and session related redux to the cafes redux feature
- Updated cafes register/deregister redux to be a persistent store of cafes and their registration state. This store holds the cafe token whenever possible
- Adds logic to refresh any expired cafe sessions when we get a node online event
- In the case of expired session refresh token, the cafe must be re-registered... This causes a query of the cafe data bundled with the app .env to get the cafe token. If it's not found there (because it's a custom cafe, for example), we consult the persistent cafe data to get the token there.


App/Models/cafes.ts
```diff
@@ -0,0 +1,20 @@
...
+const cafesString = new Buffer(cafesBase64, 'base64').toString()
+
+export const cafes: Cafe[] = JSON.parse(cafesString)
+
+export const cafesMap = cafes.reduce(
+  (acc, current) => {
+    return { ...acc, [current.peerId]: current }
+  },
+  {} as { [key: string]: Cafe | undefined }
+)
```
asutula: Making the .env bundled cafe data centrally available here

App/features/account/actions.ts
```diff
@@ -86,25 +86,3 @@ export const setRecoveryPhrase = createAction(
...
-    return (sessions: ReadonlyArray<ICafeSession>) => resolve({ sessions })
-  }
-)
-
-export const cafeSessionsError = createAction(
-  'account/CAFE_SESSIONS_ERROR',
-  resolve => {
-    return (error: any) => resolve({ error })
-  }
-)
```
asutula: Moving this cafe-related redux out of the `account` redux and into the `cafes` redux feature.

App/features/cafes/reducer.ts
```diff
@@ -1,82 +1,165 @@
...
+    readonly processing: boolean
     readonly error?: string
   }
 }
 
-export interface DeregisterCafes {
-  readonly [id: string]: {
-    readonly error?: string
-  }
-}
```
asutula: Consolidates the `RegisterCafes` and `DeregisterCafes` state into a single `Cafes` state that tracks the `registering`, `registered`, and `unregistering` state of any cafe. This data is also persistent so we can get the cafe tokens when we need them

App/features/cafes/sagas.ts
```diff
@@ -24,22 +48,99 @@ function* registerCafe(
...
+    const session = yield call(Textile.cafes.refreshSession, peerId)
+    if (session) {
+      yield put(actions.refreshCafeSession.success({ session }))
+    } else {
+      throw new Error('session not found')
+    }
+  } catch (error) {
+    const message =
+      (error.message as string) || (error as string) || 'unknown error'
+    if (message === 'unauthorized') {
```
asutula: Here's the crux of this PR, handling the case of expired refresh token

App/features/cafes/sagas.ts
```diff
@@ -24,22 +48,99 @@ function* registerCafe(
...
+      try {
+        let token: string | undefined
+        // try to get the cafe token from static cafes bundled with the app
+        const cafe = cafesMap[peerId]
+        if (cafe) {
+          token = cafe.token
```
asutula: Try to find the cafe token in the .env bundled cafe data

App/features/cafes/sagas.ts
```diff
@@ -24,22 +48,99 @@ function* registerCafe(
...
+        } else {
+          // if not, try to get the token from persisted data about registered cafes
+          const cafe: Cafe = yield select((state: RootState) =>
+            makeCafeForPeerId(peerId)(state.cafes)
+          )
+          token = cafe.token
```
asutula: Or get the token from the persisted data about registered cafes. The bundled data takes priority in case we ship updated bundled data at a later time than the cafe was registered.

App/features/cafes/sagas.ts
```diff
@@ -24,22 +48,99 @@ function* registerCafe(
...
+        yield put(actions.refreshCafeSession.failure({ peerId, error }))
+      }
+    } else {
+      yield put(actions.refreshCafeSession.failure({ peerId, error }))
+    }
+  }
+}
+
+function* refreshExpiredSessions() {
+  while (yield take([getType(TextileEventsActions.nodeOnline)])) {
```
asutula: We refresh any expired sessions as soon as the node comes online

asutula:
Ha ok, the one really is all about the logic, but I feel confident ;)

App/features/cafes/sagas.ts
sanderpick: 💎

PR #1218, opened 2019-07-21
Add session info to cafe screen

### LINKED ISSUE #1217: Add session info to cafe screen
- Display session expiration time
- Display refresh button.

undercase:
## New cafe screen

image

## Updates

- Fixes #1217.
- Adds new selector to retrieve cafe session.
- Adds expiration date and time to cafe screen.
- Adds refresh button to cafe screen to refresh session.

## Issue

The `refreshSession` method in the sdk doesn't work at the moment. It throws an error (visible in the screen above) when it is called. I investigated and I think it's just a minor issue in the sdk code-I'll consult with @asutula to fix it.


undercase:
Something I've been thinking about: I'm not sure that the peer ID display is ideal. I think the layout is good, but the information isn't conveyed super clearly. I think we should left align the peer ID and put **Peer ID:** in front of it so it's obvious what it means. Thoughts?

undercase:
Here's the error stack:

```{
  type: 'account/REFRESH_CAFE_SESSION_FAILURE',
  payload: {
    peerId: '12D3KooWGN8VAsPHsHeJtoTbbzsGjs2LTmQZ6wFKvuPich1TYmYY',
    error: {
      framesToPop: 1,
      nativeStackAndroid: [
        {
          methodName: 'fromJavaArgs',
          lineNumber: 191,
          file: 'Arguments.java'
        },
        {
          methodName: 'invoke',
          lineNumber: 32,
          file: 'CallbackImpl.java'
        },
        {
          methodName: 'resolve',
          lineNumber: 57,
          file: 'PromiseImpl.java'
        },
        {
          methodName: 'onComplete',
          lineNumber: 76,
          file: 'CafesBridge.java'
        },
        {
          methodName: 'lambda$refreshSession$2',
          lineNumber: 72,
          file: 'Cafes.java'
        },
        {
          methodName: 'call',
          lineNumber: 2,
          file: null
        }
      ],
      userInfo: null,
      code: 'refreshSession',
      line: 21956,
      column: 31,
      sourceURL: 'http://10.0.3.2:8081/index.delta?platform=android&dev=true&minify=false'
    }
  }
}```

Really, it seems like the error has to do with the callback (`onComplete`) in the bridge.

File: App/Containers/Cafe.tsx [hunk 1]
```diff
@@ -56,25 +64,50 @@ const ServiceTitle: TextStyle = {
   textAlign: 'left'
 }
 
-const ButtonContainer: ViewStyle = {
+const ExpirationDate: TextStyle = {
```
asutula: Do me a favor and update this TextStyle to use the text styles defined in `styles/textStyle`. That will go a long way in making the screens you create look consistent with the rest of the app.

App/Containers/Cafe.tsx
```diff
@@ -102,8 +135,16 @@ class Cafe extends Component<Props> {
   render() {
     // Hardcoded placeholder for services UI
     const services = ['Backup', 'Inboxing']
-    const { peer } = this.props.navigation.getParam('cafeSession').cafe
-    const buttonDisabled = !this.props.error && this.props.processing
+    const cafeSession = this.props.navigation.getParam('cafeSession')
+    const { cafe, exp } = cafeSession
+    const { peer } = cafe
+    const seconds = (exp.seconds as number) * 1000 + exp.nanos / 1000000
```
asutula: There is a util function `Textile.util.timestampToDate` that will make this much cleaner.

App/Containers/Cafe.tsx
undercase: Totally, on it.

App/Containers/Cafe.tsx
undercase: That’s convenient, thanks for the tip!

asutula:
Session refresh is working for me using your branch. Not sure what that error you see is.

asutula:
Ok, little bug and fix for it:

Currently, `Cafe.tsx`has a navigation prop `cafeSession: ICafeSession`. This used to be ok because the session was never updated while on this screen. Now that you can refresh the session from this screen, the session is updated while on this screen. Since the navigation prop `cafeSession: ICafeSession` is not connected to redux, you don't see the expiration date update when you refresh the session. 

The fix would be to make the nav prop simple `peerId: string` and then use `mapDispatchToProps` to map that peer id to other information about the session that will be updated by redux.

asutula:
Oh, I should say: Session refresh is working for me on iOS, not sure about Android, will have to test now.

asutula:
As far as the peer id display... Yea I agree it's not useful. On the previous screen (the list of all registered cafes), I used the cafe url as the primary thing and the peer id as a secondary thing. You can do the same on this screen. Hopefully we'll eventually add a cafe alias property that can be displayed, but until then, url is still the best.

undercase:
Awesome, thanks for catching that bug. Fixing it now w/ other comments.

asutula:
Ok I see the problem with calling `refreshSession` on Android. It's a bug in the RN sdk. Fixing now.

asutula:
Ok @undercase update `@textile/react-native-sdk` to `3.0.2` and that refresh session bug on Android should be fixed.

undercase:
I've addressed all of your comments and updated the screen UI/styling:

image

undercase:
Is this good to merge?

PR #1227, opened 2019-07-23
Target is always null. Moving photo.target => photo.block

andrewxhill:
fixes https://github.com/textileio/photos/issues/1226

although, please review. i'm not 100% certain if i should have fixed this, or fixed the fact that target is never present in our IFiles. 

if target no longer exists, we should remove it from the IFiles model to avoid future typescript complacency


andrewxhill:
okay, confirmed. the `target` field does need to remain on IFile, however it's only used when the file is nested within another file. so this PR is gtg

asutula:
There was also a move to use `.data`, not sure if that is related @sanderpick ?

sanderpick:
- Target is the thing that the comment was about
- Data is the link to the photo/file dag

andrewxhill:
any holdup here?

PR #1235, opened 2019-07-25
File sync group status screen

asutula:
Lets us monitor file sync in verbose ui mode. Any errors are displayable by tapping the alert icon. Once something is in a error state, that is the end of the road for that item (for now anyway), so you can clear it from the redux state by tapping the 'x' icon.
Screen Shot 2019-07-25 at 18 14 22


asutula:
Should be interesting to update an existing install with this build and see all the file sync errors and stuck items that have been accumulating :/

asutula:
Once a sync group is complete (success), it disappears from this screen

Issue #1240, opened 2019-07-26
Group creation dialog is broken; not clearing name from previous creation


andrewxhill:
1. Go to the Groups screen and click "+"
2. Name a new group and create it.
3. Navigate back to the Groups screen and click "+" again

You'll see that the form already contains the name from your first Group creation. Additionally, if you change that name to something new (or keep it the same) the form wont go anywhere (i.e. no new group can be created).

undercase:
I’ll fix this today.

andrewxhill:
close it?

PR #1247, opened 2019-07-28
Fix inability to resubmit create new thread modal

undercase:
This is a quick fix for #1240. This issue was caused by a flag in the modal's state, `submitted`, which disables the new thread functionality. It is set to `true` the first time a new thread is created, and then never reset, causing the modal to silently fail. But really, I think we should address these issues:

- Should this really be a modal? Since several other screens need to navigate to this page, and it is pretty complex, I think it might make sense to refactor it into its own navigable screen. That would simplify solving some of the problems below.
- Most of the redux state uses an indexed type for pending requests (like adding or removing a contact, ignoring a block, etc.) so that multiple requests can be theoretically handled at once. It appears that the new thread functionality only supports one new thread at a time, which makes the logic simpler, but confusing considering that it is theoretically possible to have multiple pending new thread requests at once, and other parts of the application have already solved the issue of multiple requests pretty nicely.
- Right now, the new thread modal is automatically exiting when the new thread request is submitted. So there's nowhere to display errors/issues, and the user can theoretically get navigated to the new thread at any time even after they've closed the modal (when the saga completes). Instead of passing `navigateTo` as a boolean to the action/saga, we should really support adding a callback that resets the modal/closes it upon completion, as we do for another modal in the application. Making this into an indexed screen instead of a modal would make solving this problem a lot easier and make a callback (which feels dirty to me since `redux-saga` was created to get rid of them) unnecessary.

Hope this makes sense.

### DIFF
--- a/App/Components/CreateThreadModal.tsx
+++ b/App/Components/CreateThreadModal.tsx
@@ -49,15 +49,21 @@ type Props = DispatchProps & ScreenProps
 
 interface State {
   value: string
-  submitted: boolean
 }
 
 class CreateThreadComponent extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props)
     this.state = {
-      value: props.defaultName ? props.defaultName : '',
-      submitted: false
+      value: props.defaultName ? props.defaultName : ''
+    }
+  }
+
+  componentdidUpdate() {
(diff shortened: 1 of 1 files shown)


undercase:
I also modified the code to disable the new thread modal when a request is pending to avoid the edge case mentioned above, which the `submitted` variable was intended to address.

PR #1263, opened 2019-07-31
Better photo file handling

### LINKED ISSUE #1075: Camera roll sync: Queried photos are deleted on app re-start
I saw an instance where there was a error processing some camera roll photos. They never completed the process of being added to the Textile node. They should continue to exist in a pending or error state for retry. Later on, after the app was re-started, the image files that were included in the camera roll query results no longer existed on the filesystem. I'm guessing they are being written to a tmp location. We'll need to move them to a permanent location (permanent until we explicitly delete them), either in the camer roll library or in the Photos app.

asutula:
Fixed the problem where source files for photos being added to a thread were deleted on app restart.

I also went through the flow of things and made sure that we're properly cleaning up photo source files after a photo is added successfully or anywhere sharing a photo is canceled.

Fixes #1075


App/util/copy-photo.ts
```diff
@@ -0,0 +1,31 @@
+import { call } from 'redux-saga/effects'
+import RNFS from 'react-native-fs'
+
+import AppConfig from '../Config/app-config'
+
+interface PathContaining {
+  path: string
+}
+
+export default function* copyPhoto<T extends PathContaining>(obj: T) {
```
asutula: Check it out @undercase -- Generics FTW!

undercase: Love this!

PR #1267, opened 2019-08-01
Add ability to double tap to like photos

### LINKED ISSUE #1229: Double tap to like a photo
Not sure if this would mean, remove the tappable heart or just leave both avenues

undercase:
- Fixes #1229.
- Uses the solution detailed in [this handlesbar labs article](https://medium.com/handlebar-labs/instagram-style-double-tap-with-react-native-49e757f68de), which involves wrapping the `onPress` function and using an instance variable to keep track of presses.
- Introduces a 300ms delay after an individual tap on a photo before the selected view opens up, so that if the user tags again within that time window a like will be initiated and the selected view won't open.
- Uses `setTimeout` to implement the above functionality. This is sort of dirty, but it seems like the best practice in the community based on my research.


undercase:
Ah, I didn't realize I was reinventing the wheel 💀. Updating this to use `react-native-gesture-handler` now.

undercase:
Updated to use `react-native-gesture-handler`.

asutula:
Gonna merge this since you're out @undercase

PR #1272, opened 2019-08-04
Firebase integration

asutula:
This migration to firebase turns out mostly to be adding firebase to our project so we can use it's analytics capabilities. At this time, the 'migration' of crashlytics is done through back end sync of crash data from fabric to firebase... The crashlytics library setup says in place for now, but the crash data does show up in firebase, which is nice.


.gitignore
```diff
@@ -154,5 +154,9 @@ sourcemap.js
 .jest/
 coverage/
 
+# Firebase config
+GoogleService-Info.plist
+google-services.json
```
asutula: Leaving the credentials of out git. These files will be written during CI, and we'll each have to download the dev versions for local development.

App/Sagas/index.ts
```diff
@@ -97,9 +101,23 @@ function* waitForRehydrate() {
...
 
+function* logFirebase(event: string) {
+  yield call([Firebase.analytics(), 'logEvent'], event)
+}
+
+function* subscrbeActions() {
+  while (true) {
+    const action: RootAction = yield take('*')
+    if (action.type === getType(TextileEventsActions.nodeStarted)) {
+      yield fork(logFirebase, 'tex_node_started')
```
asutula: Logging a little test event for now. Just want to see it work.

PR #1278, opened 2019-08-06
attempts to deregister and register in a catch

andrewxhill:
This is part I of III.

I went ahead and tried the 'reregister' with new cafe step because I didn't see any downside. worse case is that it fails silently and my next work will still matter. best case is that my next work wont matter (notify users of need to register a cafe).

### DIFF
--- a/App/features/cafes/sagas.ts
+++ b/App/features/cafes/sagas.ts
@@ -135,15 +135,15 @@ function* refreshExpiredSessions() {
 }
 
 function* migrateUSW() {
-  // Old us-west
-  const usw = '12D3KooWSsM117bNw6yu1auMfNqeu59578Bct5V4S9fWxavogrsw'
-  // New us-west
-  const repl = '12D3KooWSdGmRz5JQidqrtmiPGVHkStXpbSAMnbCcW8abq6zuiDP'
-  const cafes: Cafes = yield select((state: RootState) => state.cafes.cafes)
-  const peerIDs = Object.keys(cafes)
-
-  if (peerIDs.indexOf(usw) > -1) {
-    try {
+  try {
+    // Old us-west
+    const usw = '12D3KooWSsM117bNw6yu1auMfNqeu59578Bct5V4S9fWxavogrsw'
+    // New us-west
+    const repl = '12D3KooWSdGmRz5JQidqrtmiPGVHkStXpbSAMnbCcW8abq6zuiDP'
+    const cafes: Cafes = yield select((state: RootState) => state.cafes.cafes)
+    const peerIDs = Object.keys(cafes)
+  
+    if (peerIDs.indexOf(usw) > -1) {
       // Use the existing route to deregister the usw cafe
       yield put(actions.deregisterCafe.request({ peerId: usw }))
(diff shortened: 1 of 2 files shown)


App/features/cafes/sagas.ts
```diff
@@ -134,13 +134,42 @@ function* refreshExpiredSessions() {
...
 export default function*() {
   yield all([
     call(onNodeStarted),
     takeEvery(getType(actions.registerCafe.request), registerCafe),
     takeEvery(getType(actions.deregisterCafe.request), deregisterCafe),
     call(getCafeSessions),
     call(refreshExpiredSessions),
-    takeEvery(getType(actions.refreshCafeSession.request), refreshCafeSession)
+    takeEvery(getType(actions.refreshCafeSession.request), refreshCafeSession),
+    takeEvery(getType(actions.getCafeSessions.success), migrateUSW)
```
andrewxhill: i guess only question... if someone doesn't open the app for 3 months and they have expired sessions, this may never fire?

andrewxhill: nevermind, fixed that to kick off by request not by success of get sessions

asutula:
Hmm could be one of these random failures we see. Just re-launched the workflow.

andrewxhill:
that worked

App/features/cafes/sagas.ts
```diff
@@ -134,13 +134,41 @@ function* refreshExpiredSessions() {
   }
 }
 
+function* migrateUSW() {
+  try {
+    // Old us-west
+    const usw = '12D3KooWSsM117bNw6yu1auMfNqeu59578Bct5V4S9fWxavogrsw'
+    // New us-west
+    const repl = '12D3KooWSdGmRz5JQidqrtmiPGVHkStXpbSAMnbCcW8abq6zuiDP'
+    const cafes: Cafes = yield select((state: RootState) => state.cafes.cafes)
```
asutula: Sorry for chiming in after this has been merged, but I see a problem here. This saga is being called on `actions.getCafeSessions.request`. So at the time it's called, the async call to get the cafe data is in flight, but your're querying the redux state for the cafes here. It seems like that state won't be populated until a `actions.getCafeSessions.success` happens.

andrewxhill: but will cafes be populated if sessions are invalid? i had a mini-thread about this with myself on here yesterday and it's why i changed it from success => request, my concern was someone coming online in 2 months and having invalid sessions but still attempting to use the old us-west

andrewxhill: okay, if my hypothesis above is right, and the cafes object isn't populated without a session, then i can configure this to run with success or failure of sessions

asutula: I don't think the node knows anything about if the sessions are valid or not. 

The way this works now, I don't see you'd ever detect any cafe session. Seems like they'd always be empty since that request is in flight. It would only work on a subsequent run of the saga, not the first one. 

You can also just remove the dependency on other redux and sagas: This saga could run once when the main saga is created and just call the `Textile.cafes.sessions` directly to get the cafe sessions.

PR #1281, opened 2019-08-06
moves wiring to just check usw on each online

andrewxhill:


andrewxhill:
i couldn't use success or failure of the sessions very cleaning, and this model for checking cafes is already being used to look for expired cafes, so adopted i here

App/features/cafes/sagas.ts
```diff
@@ -135,29 +135,31 @@ function* refreshExpiredSessions() {
...
-          yield put(
-            actions.registerCafe.request({ peerId: repl, token: cafe.token })
-          )
+  while (yield take([getType(TextileEventsActions.nodeOnline)])) {
+    try {
+      // Old us-west
+      const usw = '12D3KooWSsM117bNw6yu1auMfNqeu59578Bct5V4S9fWxavogrsw'
+      // New us-west
+      const repl = '12D3KooWSdGmRz5JQidqrtmiPGVHkStXpbSAMnbCcW8abq6zuiDP'
+      const cafes: Cafes = yield select((state: RootState) => state.cafes.cafes)
```
asutula: Just because the node is online doesn't guarantee that redux will contain cafe session data. I'd just use this here:
```
const list: ICafeSessionList = yield call(Textile.cafes.sessions)
```

andrewxhill: sorry, confusing. to be clear. so the sessions are stored in `cafes` only if they exist but the `sessions` will store cafe details no matter what?

andrewxhill: oh, i see what you are saying

andrewxhill: maybe

PR #1286, opened 2019-08-07
Changes manifest setting from singleInstance => singleTask

andrewxhill:
fixes https://github.com/textileio/photos/issues/1041
fixes https://github.com/textileio/photos/issues/1141
fixes https://github.com/textileio/photos/issues/1287

Not going to lie, I don't fully understand this one. 

### Recreating the bug

I used the final clue in the ticket above and installed another gallery app that would take over the intent (Simple Gallery). Then when selecting an image, I'm offered selecting from Photos or Simple Gallery. If I select one and then select an image to share, `ImagePicker.launchImageLibrary` callback will return with a `user cancelled` value. 

### Solution

After looking for other people with similar issues, I found an issue I was reading through (not really related https://github.com/react-native-community/react-native-image-picker/issues/164) and noticed that there are problems if you use `android:launchMode="singleInstance"` in your manifest. I changed it and voila, it works. 

My guess is this is also the cause of the store reviewer they likely just got through onboarding fine because they took a new avatar from their camera (no intent mess up) and then just couldn't post an image because of this bug. 

Anyway, fixed.


andrewxhill:
yep, did pretty thorough QA this afternoon and found some things. 

- deep linking is working as expected now. 
- photo adds are working as expected now
- photo re-shares were broken, there was a bug in uiredux now fixed. working as expected
- fixes navigation to threads after joins now.

only odd thing left i found, is that the react-navigation params set during componentDidMount sometimes lag behind the screen. meaning if a user clicked them immediately it would error. i fixed this by disabling them until they are `!undefined`. this was actually the case before too, but I believe our app was recycling existing views so often the old values were still available, possibly causing edge-case bugs, not sure.

PR #1295, opened 2019-08-09
removed the 'add comment' view component

andrewxhill:
I was just going to change the color of an icon, but I coudn't stand looking at that super ugly comment screen anymore so got rid of it. Now it just works like this, (right to left weirdly)

Image 2

obvs can clean that up going forward, but it's better


andrewxhill:
pushing this out to you guys to try... we can change anything else downstream.

sanderpick:
Hmm, I think it should go into the message box like iMessage if there’s not going to be a second screen. Thinking about multi select too. You have to guess to know to remove the photo by retapping it.

andrewxhill:
100% on both. just wanted to start the progress, feels like a lot of this has been static for 6+ months.

sanderpick:
Tots 👍🏼

andrewxhill:
it's an intermediate that is still 10x better than the add caption screen where you could optionally change which thread you wanted to add it too, that was a relic of when we had a single group feed

PR #1301, opened 2019-08-10
Adds react-native-keyboard component and embeds a photo picker

### LINKED ISSUE #1297: Redesign the Group screen footer area (add comment/photo)
I did a rough first pass today to get rid of the extra 'addCaption' screen, but the whole footer should now be redesigned to support the new flow (complete), multi-image shared, and timeline cleanup,

- [ ] https://github.com/textileio/photos/issues/794
- [ ] https://github.com/textileio/photos/issues/1228
- [ ] https://github.com/textileio/photos/issues/1251

andrewxhill:
fixes #1297 
fixes https://github.com/textileio/photos/issues/1291

Okay, pushing forward on this. Still many mods to make about layout etc, but this solves the major bit of getting removing the need to pop out to the full screen component right away. 

After clicking the image icon: 

picker

Selecting

selected

or just typing

message


App/Components/KeyboardResponsiveContainer.tsx
```diff
@@ -4,11 +4,13 @@ import {
...
   View,
   ViewStyle,
-  LayoutAnimation
+  LayoutAnimation,
+  Platform
 } from 'react-native'
 
 interface Props {
   style?: ViewStyle
+  ios?: boolean
```
andrewxhill: this view is used in 2 places. one where i need it not to run when on iOS, this was the second simplest way to handle it. the most simple hit s TSC error, so wasn't that simple actually

App/Config/DebugConfig.ts
```diff
@@ -12,7 +12,7 @@ const config: IDebugConfig = {
   showDevScreens: __DEV__,
   useFixtures: false,
   ezLogin: false,
-  yellowBox: __DEV__,
+  yellowBox: false,
```
andrewxhill: so much nicer to develop without this... fine to revert but seems like if you want it you could toggle it on instead of having to hit 'dismiss' every time you refresh the UI

andrewxhill: firebase added a few more lines of yellow warnings here too, maybe worth signing off on @asutula

android/app/src/main/java/com/textile/textilenode/TextileImageView.java
```diff
@@ -19,7 +21,7 @@
     private int forMinWidth;
     private ScaleType scaleType;
     private boolean needsRenderScaleType = false;
-    private boolean needsRenderImage = false;
+    private boolean needsRenderImage = true;
```
andrewxhill: such a hard time getting the white thumbnails solved. however, i noticed they are prone to display on the refresh after an app crash. it gave me the hunch that maybe a rerender wasn't being requested. i thought setting this to the default, needs rerender could be good. ideally, we could postpone line 75 to set it to false only after the bitmap has been successfully rendered (since it looks like 74 is async). this is a start though, default needs a rerender... not certain the implications of that though so think it through @asutula

andrewxhill: further testing and this at least solves white thumbnails for the instance where I can create them,

1. Open app
2. Do something that causes a JS error (in my case, i just did, https://github.com/textileio/photos/issues/1213#issuecomment-520177359)
3. Reload app.

With `needsRenderImage = false` thumbnails in the main view will 100% appear as white thumbs, but with `needsRenderImage = true` they will render fine.

PR #1309, opened 2019-08-20
Notifications, cafes, storage bots

andrewxhill:
In convo with @asutula last week we decided to make better use of in-app notifications (Last screen in drawer) to instruct the user of needed input or new opportunities.

One example is Cafe registration. We can push an Alert to the notifications screen whenever a user doesn't have, deregistered, or somehow lost a cafe. Here's my plan and progress to share now for feedback.

Here's what I've done here: 

### Start highlighting unread messages

I've already added a notification visual to the app in the form of a red dot over a user's own avatar, and a red highlight notifications button in the drawer. This appears if the user has any 'unread' messages or 'alerts' (more below on alerts).

IMG_4329

### Add local alerts for things that aren't textile notifications

To do this, I moved the Notifications redux/sagas etc into their own Feature set (following the recent pattern of Groups, Contacts, and others). 

I then added a second type alongside Textile notifications, just LocalAlerts. The first alert I've created is for no cafe registered. 

### Change the Feed display to show LocalAlerts

The first LocalAlert is more complicated than just saying in one line "Do this thing", instead I wanted the ability to explain to the user why they are registering a cafe and what it adds to their experience. I also wanted to make it clear what the cafe could/couldn't do with their data. 

So I created a new card type that could be used to share this more detailed information. I riffed off the app store design (left)

IMG_4333

Other Alerts we can add here,

* invite / import contacts
* enable camera roll
* new bots

### Playing with bot name

In writing up that card, I thought it would be fun to just swap in Bot for cafe in the existing app. 

IMG_4339

Seems harmless enough that we can move forward with this and use it to help define what we'd need next to expose more options, let other expose their options...


asutula:
This is awesome :)

PR #1313, opened 2019-08-21
Updates register cafe step

andrewxhill:
because cafe.url was not part of the existing Config model, I've added a method to just get it from the gateway, now just using the local config to pull in the token when needed. Will add a few comments here too.


sanderpick:
I figured URL must be available since it’s displayed on the cafes screen. What does gateway mean here? The remote gateway?

andrewxhill:
It's known _after_ registration via peerId. Looks like config has name, peerId, token. So, I'm just grabbing from here, `https://gateway.textile.cafe/cafes`, for production. For beta/dev it's known, but i still do a check to make sure it's working the same.

andrewxhill:
For dev && beta, even though already known, I just do two checks to `https://us-west-dev.textile.cafe` and `https://us-west-beta.textile.cafe`

sanderpick:
Can you should just change the config to include url and remove peerID? I can tell you what the URLs are. They’re also in 1Password. That gateway endpoint doesn’t contain the new production cafes.

andrewxhill:
Ah, yeah, we can do that then. Will reconfig.

andrewxhill:
k, that should be gtg now

PR #1317, opened 2019-08-23
WIP: multipeer obj-c / react-native components

andrewxhill:
Pushing this up now for discussion. Maybe it would be worth cleaning up and wrapping as a small RN lib, eventually we could add the google equivalent and possibly cross-platform ble support.

### DIFF
--- a/ios/NativeModules/Multipeer.m
+++ b/ios/NativeModules/Multipeer.m
@@ -1,0 +1,88 @@
+#import "Multipeer.h"
+
+@interface Multipeer()
+
+@end
+
+@implementation Multipeer
+
+RCT_EXPORT_MODULE();
+
+RCT_EXPORT_METHOD(init:(NSString *)name address:(NSString *)address avatar:(NSString *)avatar) {
+  [self initialize:name address:address avatar:avatar];
+}
+
+RCT_EXPORT_METHOD(advertiseOn) {
+  [self advertiseToggle:true];
+}
+
+RCT_EXPORT_METHOD(advertiseOff) {
+  [self advertiseToggle:false];
+}
+
+RCT_EXPORT_METHOD(scan) {
+  [self scanStart];
+}
+
+#pragma mark -
+#pragma mark Public Methods
+- (void)initialize:(NSString *)name address:(NSString *)address avatar:(NSString *)avatar {
+  self.name = name;
+  self.address = address;
+  self.avatar = avatar;
+  self.peerID = [[MCPeerID alloc] initWithDisplayName:address];
+  self.session = [[MCSession alloc] initWithPeer:self.peerID];
+  self.session.delegate = self;
+}
+
+- (void)scanStart {
+  self.browser = [[PeerBrowser alloc] init];
+  [self.browser initialize:@"textileio" peerID:self.peerID];
+}
+
+- (void)advertiseToggle:(BOOL)advertise {
+  if (advertise) {
+    NSDictionary *discoveryInfo = @{ @"address": self.address, @"avatar": self.avatar, @"name": self.name };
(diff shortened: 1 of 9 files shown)


ios/NativeModules/Multipeer.m
```diff
@@ -0,0 +1,88 @@
...
+
+@interface Multipeer()
+
+@end
+
+@implementation Multipeer
+
+RCT_EXPORT_MODULE();
+
+RCT_EXPORT_METHOD(init:(NSString *)name address:(NSString *)address avatar:(NSString *)avatar) {
```
andrewxhill: Okay, you see I init with some Textile specific variables. The idea here was that when detected by another device, that device could show the real Textile contact info.

ios/NativeModules/Multipeer.m
```diff
@@ -0,0 +1,88 @@
...
+}
+
+- (void)scanStart {
+  self.browser = [[PeerBrowser alloc] init];
+  [self.browser initialize:@"textileio" peerID:self.peerID];
+}
+
+- (void)advertiseToggle:(BOOL)advertise {
+  if (advertise) {
+    NSDictionary *discoveryInfo = @{ @"address": self.address, @"avatar": self.avatar, @"name": self.name };
```
andrewxhill: here, it adds that Textile data directly to the broadcast info. Which means other users can find all the relevant textile bits without having to make a connection.

ios/NativeModules/Multipeer.m
```diff
@@ -0,0 +1,88 @@
...
+  self.address = address;
+  self.avatar = avatar;
+  self.peerID = [[MCPeerID alloc] initWithDisplayName:address];
+  self.session = [[MCSession alloc] initWithPeer:self.peerID];
+  self.session.delegate = self;
+}
+
+- (void)scanStart {
+  self.browser = [[PeerBrowser alloc] init];
+  [self.browser initialize:@"textileio" peerID:self.peerID];
```
andrewxhill: advertising and browsing is namespaced by textile. this is cool because we could generate namespaces for say, threads, where only others already in the thread might listen for those.

ios/NativeModules/Multipeer.m
```diff
@@ -0,0 +1,88 @@
...
+  }
+}
+
+#pragma mark -
+#pragma mark Session Delegate Methods
+- (void)session:(MCSession *)session peer:(MCPeerID *)peerID didChangeState:(MCSessionState)state {
+//  NSDictionary *userInfo = @{ @"peerID": peerID,
+//                              @"state" : @(state) };
+//  dispatch_async(dispatch_get_main_queue(), ^{
+//    [[NSNotificationCenter defaultCenter] postNotificationName:@"Multipeer_DidChangeStateNotification"
```
andrewxhill: didn't bother with the events yet, was just using breakpoints to ensure i was getting my two phones chit chatting

ios/NativeModules/Multipeer.m
```diff
@@ -0,0 +1,88 @@
...
+  [self advertiseToggle:false];
+}
+
+RCT_EXPORT_METHOD(scan) {
+  [self scanStart];
+}
+
+#pragma mark -
+#pragma mark Public Methods
+- (void)initialize:(NSString *)name address:(NSString *)address avatar:(NSString *)avatar {
```
andrewxhill: coudn't tell the energy hit. but seems like some cool things we could test.

1. Invite by Multipeer. So in a room, just tap their icons and send the invite payload over p2p. More or less we have this already with the shared deeplinks.
2. I think cooler could be to test a 'Create a Local Group' that just opens up the invite to anyone listening, and when they join they also open up the invite, etc etc.

ios/NativeModules/Multipeer.m
andrewxhill: if we did a "Local Group" this is also where we could just add the current group

